PR #7, opened 2018-04-04
Feature/tests

Henrod:
# What changed
* new error on pomelo packet decoder
* tests for pomelo packet decoder


internal/codec/pomelo_packet_decoder_test.go
```diff
@@ -0,0 +1,82 @@
...
+	buf []byte
+	err error
+}{
+	"test_ok":                   {[]byte{packet.Handshake}, nil},
+	"test_wrong_packet_type":    {[]byte{0x06}, packet.ErrWrongPomeloPacketType},
+	"test_over_max_packet_size": {[]byte{packet.Handshake, 0xFF, 0xFF, 0xFF}, ErrPacketSizeExcced},
+}
+
+var (
+	// headers
```
felipejfc: this comment seems unnecessary

internal/codec/pomelo_packet_decoder_test.go
```diff
@@ -0,0 +1,82 @@
...
+	handshakeHeader = []byte{packet.Handshake, 0x00, 0x00, 0x01}
+	invalidHeader   = []byte{0x06, 0x00, 0x00, 0x01}
+)
+
+var decodeTables = map[string][]struct {
+	data   []byte
+	packet []*packet.Packet
+	err    error
+}{
+	"test_not_enough_bytes":       {{[]byte{0x01}, nil, constants.ErrMessageWithNotEnoughLength}},
```
felipejfc: it would be nice (and also seems easy) to create a case that would throw ErrPacketSizeExcced

felipejfc: also, create cases that cover all packet.Type on https://github.com/topfreegames/pitaya/blob/master/internal/packet/constants.go

Henrod: I didn't created a test for ErrPacketSizeExcced because it is already tested on TestForward. So on TestDecode, it assumes forward is correct (and tested) and just returns some error to fall on this error case on Decode.

Henrod: Ok, makes sense to create a test case for each packet.Type

felipejfc:
its good now but you need to rebase with master so that we can see the tests working and then I'll merge

PR #10, opened 2018-04-23
Instead of returning an error, panic when the number of certificates …

leohahn:
…is incorrect passed to the acceptor.


felipejfc:
thanks!

PR #12, opened 2018-05-04
log heartbeat if true on config

Henrod:


felipejfc:
you are missing a default value, I recommend setting default true in config.go

Henrod:
Actually I thought that would be better if the default value was false (that's why I didn't put on config.go). If the heartbeat fails, an error would be logged and we would know.

felipejfc:
I do, this is useful information and that's what debug log is for...

PR #14, opened 2018-05-18
Fix nats rpc server shutdown

andrehp:


felipejfc:
why is that needed?

andrehp:
Pitaya crashes if the nats server receives a message after the shutdown command

felipejfc:
I see, makes sense then

Em sex, 18 de mai de 2018 às 16:00, Andre Hahn Pereira <
notifications@github.com> escreveu:

PR #16, opened 2018-05-21
fatal when disconnected from nats after retries

Henrod:
Now the max number of reconnection tries on Nats is configurable with `pitaya.cluster.rpc.client.nats.maxReconnects` (default is 15, with one try every 2s).

Also, when nats connection is closed, it logs a Fatal and exits the process. The nats connection is closed in two situations: when conn.Close() is called or when nats client tried to reconnect to nats server more than maxReconnections. 

With the log.Fatal, I made a hacky test and I want to check if it is ok.

When the client disconnects from Nats after retries, is uses log.Fatal, which logs an error and executes os.Exit(1). 

The hacky part is to, inside the test, execute a command that run the same test with env var STOP_CONN=true, which calls conn.Close(). Then, in the first test, I check that the process exited with code 1.


cluster/nats_rpc_client.go
```diff
@@ -46,6 +46,7 @@ type NatsRPCClient struct {
 	config          *config.Config
 	conn            *nats.Conn
 	connString      string
+	maxReconnects   int
```
felipejfc: maxReconnectionRetries

cluster/nats_rpc_client.go
```diff
@@ -71,6 +72,7 @@ func (ns *NatsRPCClient) configure() error {
 	if ns.connString == "" {
 		return constants.ErrNoNatsConnectionString
 	}
+	ns.maxReconnects = ns.config.GetInt("pitaya.cluster.rpc.client.nats.maxreconnects")
```
felipejfc: pitaya.cluster.rpc.client.nats.maxreconnectionretries

cluster/nats_rpc_client_test.go
```diff
@@ -364,7 +365,36 @@ func TestNatsRPCClientCall(t *testing.T) {
 			assert.Equal(t, table.err, err)
 			err = subs.Unsubscribe()
 			assert.NoError(t, err)
-			conn.Close()
+			// conn.Close()
```
felipejfc: 🤭

cluster/nats_rpc_client_test.go
```diff
@@ -364,7 +365,36 @@ func TestNatsRPCClientCall(t *testing.T) {
...
 		})
 	}
 }
+
+func TestCloseNatsClient(t *testing.T) {
+	// TODO: is this too hacky?
```
felipejfc: It is!

PR #18, opened 2018-06-01
Use protobuf only for RPC calls

felipejfc:
This will make it possible to have servers made in other programming languages to send and receive RPC calls to go servers.

- Remotes should now receive and return structs that implements proto.Message interface
- session.Data and context are now json encoded instead of gob for sending in RPC calls


protos/pitaya.proto
```diff
@@ -52,3 +52,16 @@ message Error {
   string msg = 2;
   map<string, string> metadata = 3;
 }
+
+message KickMsg {
+  string UserID = 1;
```
cscatolini: userId

protos/pitaya.proto
```diff
@@ -52,3 +52,16 @@ message Error {
...
+}
+
+message KickAnswer {
+  bool Kicked = 1;
```
cscatolini: kicked

protos/pitaya.proto
```diff
@@ -52,3 +52,16 @@ message Error {
...
+}
+
+message BindMsg {
+  string UID = 1;
```
cscatolini: uid, fid

rpc.go
```diff
@@ -24,21 +24,22 @@ import (
...
 
+	"github.com/gogo/protobuf/proto"
 	"github.com/topfreegames/pitaya/constants"
 	"github.com/topfreegames/pitaya/route"
 )
 
 // RPC calls a method in a different server
-func RPC(ctx context.Context, routeStr string, reply interface{}, args ...interface{}) error {
-	return doSendRPC(ctx, "", routeStr, reply, args...)
+func RPC(ctx context.Context, routeStr string, reply proto.Message, arg proto.Message) error {
```
cscatolini: we no longer accept RPC calls with several arguments?

PR #22, opened 2018-06-26
send default tags on reporter

Henrod:
I would like to choose new tags (or labels, on Prometheus) to send to StatsD and Prometheus. This PR adds a new field on config to get these default tags and sends them on every metric.

### DIFF
--- a/app.go
+++ b/app.go
@@ -137,13 +137,13 @@ func Configure(
 	app.messageEncoder = message.NewMessagesEncoder(app.config.GetBool("pitaya.handler.messages.compression"))
 	app.configured = true
 	app.metricsReporters = make([]metrics.Reporter, 0)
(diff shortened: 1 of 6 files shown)


app.go
```diff
@@ -137,13 +137,13 @@ func Configure(
 	app.messageEncoder = message.NewMessagesEncoder(app.config.GetBool("pitaya.handler.messages.compression"))
 	app.configured = true
 	app.metricsReporters = make([]metrics.Reporter, 0)
-	AddMetricsReporter(metrics.GetPrometheusReporter(serverType, app.config.GetString("pitaya.game"), app.config.GetInt("pitaya.metrics.prometheus.port")))
+
+	defaultTags := app.config.GetStringMapString("pitaya.metrics.tags")
```
felipejfc: can you change the name to GetMapStringString? :)

felipejfc: changed my mind

PR #24, opened 2018-06-28
Add serializer to handshake response and save handshake data on the session.

leohahn:


service/handler.go
```diff
@@ -207,14 +209,27 @@ func (h *HandlerService) Handle(conn net.Conn) {
 }
 
 func (h *HandlerService) processPacket(a *agent.Agent, p *packet.Packet) error {
+	fmt.Println("PROCESSING PAKCET MAMN")
```
Henrod: Should this be here?

service/handler.go
```diff
@@ -207,14 +209,27 @@ func (h *HandlerService) Handle(conn net.Conn) {
...
-		a.SetStatus(constants.StatusHandshake)
 		logger.Log.Debugf("Session handshake Id=%d, Remote=%s", a.Session.ID(), a.RemoteAddr())
 
+		// Parse the json sent with the handshake by the client
+		handshakeData := &session.HandshakeData{}
+		err := json.Unmarshal(p.Data, handshakeData)
+
+		if err != nil {
+			a.SetStatus(constants.StatusClosed)
+			println("INVALID HANDSHAKE DATA BOI")
```
Henrod: Should this be here?

session/session.go
```diff
@@ -58,6 +58,22 @@ var (
 	SessionCount int64
 )
 
+// HandshakeClientData represents information about the client sent on the handshake.
+type HandshakeClientData struct {
+	Platform    string `json:"platform"`
+	LibVersion  string `json:"lib_version"`
```
Henrod: I think the json format should be libVersion (camel case). But if the client also is lib_version, that's not a big problem.

session/session.go
```diff
@@ -69,6 +85,7 @@ type Session struct {
 	lastTime          int64                  // last heartbeat time
 	entity            NetworkEntity          // low-level network entity
 	data              map[string]interface{} // session data store
+	handshakeData     *HandshakeData         // handshake data received by the client
```
Henrod: Why didn't you put the handshake data on session's data?

service/handler.go
leohahn: Oops, will remove that.

session/session.go
leohahn: It can be camel case as well. I can change it.

service/handler.go
cscatolini: best log 🤣

session/session.go
leohahn: That way the user would have to keep in mind of the handshake data when changing the session data. For example, clearing the session data would also clear the handshake data, this not being very obvious.

PR #27, opened 2018-07-16
Add new field GetName on serializer interface

leohahn:


agent/agent_test.go
```diff
@@ -76,6 +76,7 @@ func TestNewAgent(t *testing.T) {
 	hbTime := time.Second
 
 	mockConn := mocks.NewMockConn(ctrl)
+	mockSerializer.EXPECT().GetName().Times(1)
```
Henrod: When expect only once, there is no need to add .Times(1)

service/handler_test.go
```diff
@@ -249,6 +253,8 @@ func TestHandlerServiceProcessPacketHandshake(t *testing.T) {
 			defer ctrl.Finish()
 
 			mockSerializer := serializemocks.NewMockSerializer(ctrl)
+			mockSerializer.EXPECT().GetName().AnyTimes()
```
Henrod: AnyTimes is not very good. To be sure it is called once, just drop the .AnyTimes()

PR #29, opened 2018-07-19
Make prometheus monitoring optional

andrehp:


config/config.go
```diff
@@ -89,6 +89,7 @@ func (c *Config) fillDefaultValues() {
 		"pitaya.metrics.statsd.rate":                            1,
 		"pitaya.metrics.prometheus.port":                        9090,
 		"pitaya.metrics.tags":                                   map[string]string{},
+		"pitaya.metrics.prometheus.enabled":                     true,
```
felipejfc: I'm thinking of changing this default value to false and logging that monitoring is disabled... this would make development easier IMO

andrehp: I left it enabled by default to avoid breaking existing applications, but I agree that leaving it disabled by default would avoid some pain when running locally

felipejfc: we just need to ensure to add the env var to turn it on on the current production applications and then it will be safe to turn it off by default :)

andrehp: Disabled it by default now

PR #34, opened 2018-08-08
Bugfix/sd sigint (closes #11)

### LINKED ISSUE #11: Can't exit pitaya if service discovery stops running
If service discovery stops running I cannot exit pitaya and have to manually kill the process.

gabrielerzinger:
Made two modifications to enable graceful shutdown when etcd is unavailable and exceeds timeout.

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -261,7 +261,6 @@ func (sd *etcdServiceDiscovery) bootstrap() error {
 		}
 		return nil
 	}
-
 }
 
 // GetServer returns a server given it's id
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -55,12 +55,15 @@ type etcdServiceDiscovery struct {
 	lastSyncTime        time.Time
 	listeners           []SDListener
 	revokeTimeout       time.Duration
+	bootstrapTimeout    time.Duration
+	appDieChan          chan bool
 }
 
 // NewEtcdServiceDiscovery ctor
 func NewEtcdServiceDiscovery(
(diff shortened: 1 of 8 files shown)


config/config.go
```diff
@@ -74,6 +74,8 @@ func (c *Config) fillDefaultValues() {
 		"pitaya.cluster.sd.etcd.endpoints":                      "localhost:2379",
 		"pitaya.cluster.sd.etcd.prefix":                         "pitaya/",
 		"pitaya.cluster.sd.etcd.heartbeat.ttl":                  "60s",
+		"pitaya.cluster.sd.etcd.bootstrap.timeout":              "180s",
```
andrehp: Please update the docs to include these new configuration values

cluster/etcd_service_discovery.go
```diff
@@ -232,16 +239,29 @@ func (sd *etcdServiceDiscovery) GetServersByType(serverType string) (map[string]
...
+		err := sd.bootstrapLease()
+		if err != nil {
+			c <- err
+			return
+		}
+		err = sd.bootstrapServer(sd.server)
+		c <- err
+	}()
+	select {
+	case err := <-c:
```
andrehp: Do you need to signal de application to die gracefully here? Because what happens if this function returns an error is that the startModules function in module.go will raise a fatal error and exit the application

It seems that the behavior on error or timeout should be the same. Given that both cases block the application from really starting.

gabrielerzinger: That should be fixed on a next pull request by refactoring the bootstrap/lease logic

andrehp: Ok

PR #35, opened 2018-08-08
[DONE] Fix concurrently closing and writing to the same chan

rcmgleite:
## Problem
The go client spawns two goroutines: handlePackets and pendingRequestsReaper

They both write to client.IncomingMsgChan.
The problem is: the client.Disconnect method was closing the IncomingMsgChan
and there is no way for the handlePackets and pendingRequestReaper goroutines to know 
that this happened.
When one of these 2 goroutines tries to write to the closed channel the application panics.

## Solution
I'm not completely sure that just not closing the IncomingMsgChan is the correct solution
but in some articles like [this](https://go101.org/article/channel-closing.html) there is a 'rule' to only close channels from the writer goroutine.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -301,7 +301,6 @@ func (c *Client) Disconnect() {
 	if c.Connected {
 		c.Connected = false
 		close(c.closeChan)
-		close(c.IncomingMsgChan)
 		c.conn.Close()
 	}
 }


Henrod:
That way won't you leak channels?

rcmgleite:
@Henrod is there a simple way to verify it?
I tried to monitor memory usage using the client with this fix and everything seemed normal

client/client.go
```diff
@@ -300,8 +300,6 @@ func (c *Client) sendHeartbeats(interval int) {
 func (c *Client) Disconnect() {
 	if c.Connected {
 		c.Connected = false
-		close(c.closeChan)
```
cscatolini: https://github.com/topfreegames/pitaya/blob/724c8df963caff62a5f61040cb13e38909fef3b4/client/client.go#L207

`closeChan` is actually being used to stop these goroutines:

```
	go c.handlePackets()
	go c.pendingRequestsReaper()
```

so by not closing it you will have a leak

felipejfc:
can't you solve this with a lock in both of the goroutines + the close() method?

rcmgleite:
as @cscatolini pointed out, the closeChan must be closed to avoid leaking goroutines.
As for the IncomingMsgChan, it seems that not closing it would no be a problem.
@felipejfc

felipejfc:
I guess we could leak channels by not closing them but in this case it doesn't seems like a reason for not doing it because we're only using this client for tests... (depending on how we implement pitaya-bot this may be a point of memory leak tho)

felipejfc:
After doing some research I guess it is @rcmgleite, you can merge this

PR #39, opened 2018-08-14
ordered modules

Henrod:
Initialise modules in order, in case the binary modules needs to use rpc client module for example.

I introduced the function RegisterModuleBefore, so don't force the developer to initialise and register a module before others.


andrehp:
LGTM, just need to fix the linting errors

PR #40, opened 2018-08-14
Added a method to get all servers in cluster

gabrielerzinger:
We need a func on package pitaya to get all servers


cluster/etcd_service_discovery.go
```diff
@@ -238,6 +238,25 @@ func (sd *etcdServiceDiscovery) GetServersByType(serverType string) (map[string]
 	return nil, constants.ErrNoServersAvailableOfType
 }
 
+// GetAllServers returns a slice with all the servers
+func (sd *etcdServiceDiscovery) GetAllServers() ([]*Server, error) {
+	ret := make([]*Server, 0)
+	var rangeErr error
+	sd.serverMapByID.Range(func(k, v interface{}) bool {
+		s, err := sd.GetServer(k.(string))
```
felipejfc: you dont need to getServer(k), your v is already what you are getting here

app.go
```diff
@@ -259,6 +259,11 @@ func GetServersByType(t string) (map[string]*cluster.Server, error) {
 	return app.serviceDiscovery.GetServersByType(t)
 }
 
+// GetAllServers get all servers
+func GetAllServers() ([]*cluster.Server, error) {
```
felipejfc: I prefer to call it GetServers

felipejfc:
tests are failing @gabrielerzinger

PR #41, opened 2018-08-16
[Done] Pipeline now receives the decoded arguments insted of raw data

rcmgleite:
### This PR changes the signature of the pipeline handler.
The ideia is to make possible for pipelines to access the decoded struct from the message bytes.

#### Motivation
Make it possible to, for example, create pipelines for struct field validation (see the example added to /examples/demo/pipeline)

#### TODOs
- [x] Fix examples that used pipelines with the old signature

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -180,6 +180,25 @@
   version = "v1.0.0"
 
 [[projects]]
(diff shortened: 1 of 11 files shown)


rcmgleite:
@cscatolini done!

cscatolini:
Great! Thanks! Is this still WIP? Can we merge this?

rcmgleite:
Its done!

PR #42, opened 2018-08-16
[DONE] Feature/struct validation pipeline

rcmgleite:


docs/configuration.rst
```diff
@@ -245,3 +245,21 @@ These configurations are only used if the modules are created. It is recommended
...
+  :stub-columns: 1
+
+  * - Configuration
+    - Default value
+    - Type
+    - Description
+  * - pitaya.defaultpipelines.structvalidation.enabled
+    - false
+    - bool
+    - Whether Pitaya should use the default struct validator for handler arguments
```
andrehp: I think `Whether Pitaya should enable the default struct validator for handler arguments` would sound better

PR #43, opened 2018-08-16
WIP: Kick refactor

gabrielerzinger:
[**WIP**] Refactored kick to work the same on NATS and GRPC


cluster/nats_rpc_server.go
```diff
@@ -267,6 +287,18 @@ func (ns *NatsRPCServer) processPushes() {
 	}
 }
 
+func (ns *NatsRPCServer) processKick() {
+	for kick := range ns.getUserKickChannel() {
+		k := &protos.KickMsg{}
+		err := proto.Unmarshal(kick.Data, k)
```
golangci: ineffectual assignment to `err`

kick.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/protos"
+	"github.com/topfreegames/pitaya/session"
+)
+
+// SendKickToUser sends kick to an user
+func SendKickToUser(uid, frontendType string) error {
+	if s := session.GetSessionByUID(uid); s != nil {
```
andrehp: You should check if the current server is a frontend server here, this call doesn't make sense for backend servers

kick.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"context"
+
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/protos"
+	"github.com/topfreegames/pitaya/session"
+)
+
+// SendKickToUser sends kick to an user
+func SendKickToUser(uid, frontendType string) error {
```
andrehp: It would be nice to have a function that accepts a slice of uids to kick as well

gabrielerzinger: Multiples uids from multiples servers or multiples uids from one server ? (the last is what we have on push.go)

cluster/nats_rpc_client_test.go
```diff
@@ -167,6 +167,40 @@ func TestNatsRPCClientBroadcastSessionBind(t *testing.T) {
...
+	uid := "testuid"
+	s := helpers.GetTestNatsServer(t)
+	defer s.Shutdown()
+	cfg := viper.New()
+	cfg.Set("pitaya.cluster.rpc.client.nats.connect", fmt.Sprintf("nats://%s", s.Addr()))
+	config := getConfig(cfg)
+	sv := getServer()
+
+	rpcClient, _ := NewNatsRPCClient(config, sv, nil, nil)
+	rpcClient.Init()
```
golangci: Error return value of `rpcClient.Init` is not checked

cluster/nats_rpc_client_test.go
```diff
@@ -167,6 +167,40 @@ func TestNatsRPCClientBroadcastSessionBind(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	m := helpers.ShouldEventuallyReceive(t, kickChan).(*nats.Msg)
+
+	actual := &protos.KickMsg{}
+	err = proto.Unmarshal(m.Data, actual)
+	assert.NoError(t, err)
+
+	assert.Equal(t, kick.UserId, actual.UserId)
+	subs.Unsubscribe()
```
golangci: Error return value of `subs.Unsubscribe` is not checked

cluster/nats_rpc_server_test.go
```diff
@@ -143,6 +152,24 @@ func TestNatsRPCServerSubscribeToBindingsChannel(t *testing.T) {
...
+	rpcServer, _ := NewNatsRPCServer(cfg, sv, nil, nil)
+	s := helpers.GetTestNatsServer(t)
+	defer s.Shutdown()
+	conn, err := setupNatsConn(fmt.Sprintf("nats://%s", s.Addr()), nil)
+	assert.NoError(t, err)
+	rpcServer.conn = conn
+	err = rpcServer.subscribeToUserKickChannel("someuid", sv.Type)
+	assert.NoError(t, err)
+	dt := []byte("somedata")
+	conn.Publish(GetUserKickTopic("someuid", sv.Type), dt)
```
golangci: Error return value of `conn.Publish` is not checked

kick_test.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, table := range tables {
+		t.Run(table.name, func(t *testing.T) {
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+			mockNetworkEntity := mocks.NewMockNetworkEntity(ctrl)
+
+			uid1 := uuid.New().String()
+			uid2 := uuid.New().String()
+			s1 := session.New(mockNetworkEntity, true)
+			err := s1.Bind(nil, uid1)
```
golangci: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use

kick_test.go
```diff
@@ -0,0 +1,96 @@
...
+			assert.NoError(t, err)
+
+			s2 := session.New(mockNetworkEntity, true)
+			err = s2.Bind(nil, uid2)
```
golangci: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use

kick_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	clustermocks "github.com/topfreegames/pitaya/cluster/mocks"
+	"github.com/topfreegames/pitaya/protos"
+	"github.com/topfreegames/pitaya/session"
+	"github.com/topfreegames/pitaya/session/mocks"
+)
+
+func TestSendKickToUsersLocalSession(t *testing.T) {
```
andrehp: It would be nice to add some error case scenarios as well, such as when Kick returns an error

andrehp:
Good job

cluster/nats_rpc_server.go
```diff
@@ -52,6 +52,7 @@ type NatsRPCServer struct {
 	bindingsChan           chan *nats.Msg // bindingsChan receives notify from other servers on every user bind to session
 	unhandledReqCh         chan *protos.Request
 	userPushCh             chan *protos.Push
+	userKickCh             chan *nats.Msg
```
cscatolini: Why did we change the standard here? you should probably use `protos.KickMsg` instead of `nats.Msg` and `processKick` would be more like `processPushes`. I am aware that the way it is now will work, but I'd rather keep the standards we set before. 

Since this was already merged, could you open another PR changing this, please? @gabrielerzinger

PR #44, opened 2018-08-17
[DONE] Add default logger to context on both handlers and remotes

rcmgleite:
This is an ideia: to have a default logger within each handler/remote ctx that already has some fields to help debugging.

As I'm not sure if this should be included directly on pitaya core I requested the review from all of you.

Could you please just give your opinions and possible alternatives for it?

### Goal
Have a default logger on each handler and remote with fields like: route, sessionsId, requestId

### DIFF
--- a/app.go
+++ b/app.go
@@ -502,6 +502,11 @@ func GetSessionFromCtx(ctx context.Context) *session.Session {
 	return ctx.Value(constants.SessionCtxKey).(*session.Session)
 }
 
+// GetDefaultLoggerFromCtx returns the default logger from the given context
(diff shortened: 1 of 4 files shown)


Henrod:
I like it. It is useful to always have a logger with requestId, sessionId and route name. That saves me from creating it on every request and maybe forgetting to add sessionId as a field, for example. Besides, it standardises  the route name format on logger field.

rcmgleite:
@cscatolini could you give a final look and possibly merge?

cscatolini:
@rcmgleite great job! I'll merge this :)

PR #46, opened 2018-08-21
default pitaya logger examples

rcmgleite:


rcmgleite:
@andrehp any ideas on where to add real documentation for the logger and struct validation?

PR #55, opened 2018-09-12
Additional Tags and routingFunc

Henrod:
1. Use ctx on routingFunc for it can be more flexible for development than session
2. Change config `pitaya.metrics.tags` to `pitaya.metrics.constTags` since these tags are indeed constants
3. Create config `pitaya.metrics.additionalTags` which is a map from tag to default value. The values can be added using the method `AddMetricTagsToPropagateCtx`

### DIFF
--- a/metrics/prometheus_reporter.go
+++ b/metrics/prometheus_reporter.go
@@ -233,7 +233,7 @@ func (p *PrometheusReporter) ReportGauge(metric string, labels map[string]string
 	return constants.ErrMetricNotKnown
 }
 
-// ensureLabels check if labels contains the additionalLabels values,
+// ensureLabels checks if labels contains the additionalLabels values,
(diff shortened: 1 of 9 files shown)


metrics/prometheus_reporter.go
```diff
@@ -211,8 +226,21 @@ func (p *PrometheusReporter) ReportCount(metric string, labels map[string]string
...
 	g := p.gaugeReportersMap[metric]
 	if g != nil {
+		labels = p.ensureLabels(labels)
 		g.With(labels).Set(value)
 		return nil
 	}
 	return constants.ErrMetricNotKnown
 }
+
+// ensureLabels check if labels contains the additionalLabels values,
```
andrehp: s/check/checks

andrehp:
Good job

PR #56, opened 2018-09-14
Push and kick fix ( Closes #51 ) 

### LINKED ISSUE #51: SendPushToUsers possibly not returning error if last push was successfull
I think that `SendPushToUsers` on the main package may have the following problem:

If an array of N users get success on pushing the message for the last user, it will always return `nil`. This can lead the game to believe that the push was successful while it may have failed for one or more users. If this is intentional, I think maybe we should document it somehow.

This problem may also exist on `SendKickToUsers` since this one was based on the first and I didnt saw that early, and im going to fix kick asap. But for the push one, i would like know better if this is intentional or not

gabrielerzinger:
This pull request address to issue #51 with a possible solution.

It is a WIP because it **still needs to be rebased with master** **and** because it will break compatibility with projects that uses Pitaya, therefore, we need to version it before. Although, i've opened the PR so we can discuss if this solution is appropriate.


andrehp:
Please fix the conflicts

andrehp:
I believe the solution is good and lets the applications check which user ids failed, we just need to fix some things and then we are ready to go.

andrehp:
Good job, let's release this as v0.3.0

PR #58, opened 2018-09-18
Feature/grpc out cluster

Henrod:
Enable grpc between clusters 
- graceful shutdown on grpc servers
- server publish external host and region so other servers decide which host to use

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -61,6 +61,13 @@ type RemoteBindingListener interface {
 	OnUserBind(uid, fid string)
 }
 
+// InfoRetriever gets cluster info
+// It can be implemented, for exemple, by reading
+// env var, config or by accessing the cluster API
+type InfoRetriever interface {
+	Region() string
+}
+
 // Action type for enum
 type Action int
 
--- a/cluster/config_info_retriever.go
+++ b/cluster/config_info_retriever.go
@@ -1,0 +1,21 @@
+package cluster
+
+import "github.com/topfreegames/pitaya/config"
+
+// ConfigInfoRetriever gets cluster info from config
+// Implements InfoRetriever interface
+type ConfigInfoRetriever struct {
+	region string
+}
+
+// NewConfigInfoRetriever returns a *ConfigInfoRetriever
+func NewConfigInfoRetriever(c *config.Config) *ConfigInfoRetriever {
+	return &ConfigInfoRetriever{
+		region: c.GetString("pitaya.cluster.info.region"),
+	}
+}
+
+// Region gets server info from env var
+func (c *ConfigInfoRetriever) Region() string {
+	return c.region
+}
--- a/cluster/config_info_retriever.go
+++ b/cluster/config_info_retriever.go
@@ -15,7 +15,7 @@ func NewConfigInfoRetriever(c *config.Config) *ConfigInfoRetriever {
 	}
 }
 
-// Region gets server info from env var
+// Region gets server's region from config
 func (c *ConfigInfoRetriever) Region() string {
 	return c.region
 }
--- a/cluster/grpc_rpc_client.go
+++ b/cluster/grpc_rpc_client.go
@@ -48,21 +48,25 @@ type GRPCClient struct {
 	metricsReporters []metrics.Reporter
 	clientMap        sync.Map
 	bindingStorage   interfaces.BindingStorage
+	infoRetriever    InfoRetriever
 	reqTimeout       time.Duration
 	dialTimeout      time.Duration
 }
 
 // NewGRPCClient returns a new instance of GRPCClient
-func NewGRPCClient(config *config.Config, server *Server, metricsReporters []metrics.Reporter, bindingStorage ...interfaces.BindingStorage) (*GRPCClient, error) {
-	var bs interfaces.BindingStorage
-	if len(bindingStorage) > 0 {
-		bs = bindingStorage[0]
-	}
+func NewGRPCClient(
+	config *config.Config,
+	server *Server,
+	metricsReporters []metrics.Reporter,
+	bindingStorage interfaces.BindingStorage,
+	infoRetriever InfoRetriever,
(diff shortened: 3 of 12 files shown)


cluster/config_info_retriever.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// NewConfigInfoRetriever returns a *ConfigInfoRetriever
+func NewConfigInfoRetriever(c *config.Config) *ConfigInfoRetriever {
+	return &ConfigInfoRetriever{
+		region: c.GetString("pitaya.cluster.info.region"),
+	}
+}
+
+// Region gets server info from env var
```
andrehp: The info comes from the config

cluster/cluster.go
```diff
@@ -61,6 +61,13 @@ type RemoteBindingListener interface {
 	OnUserBind(uid, fid string)
 }
 
+// InfoRetriever gets cluster info
```
andrehp: Is it cluster or region info?

cluster/grpc_rpc_client.go
```diff
@@ -201,14 +205,18 @@ func (gs *GRPCClient) SendPush(userID string, frontendSv *Server, push *protos.P
...
+
+	host = gs.getServerHost(sv)
+	if host == "" {
+		logger.Log.Errorf("server %s has no grpc-host specified in metadata", sv.ID)
 		return
 	}
-	if port, ok = sv.Metadata["grpc-port"]; !ok {
-		logger.Log.Errorf("server %s doesn't have a grpc-port specified in metadata", sv.ID)
+
+	if port, ok = sv.Metadata[constants.GRPCPortKey]; !ok {
```
andrehp: Does the internal port need to be the same as the external one?

cluster/cluster.go
Henrod: The intension is to be extendable to new values, not only region

cluster/grpc_rpc_client.go
Henrod: Two different ports wouldn't add an unnecessary complexity to the configuration?
But maybe it's better to be more flexible and accept different ports

cluster/config_info_retriever.go
Henrod: 😱

cluster/grpc_rpc_client.go
andrehp: I feel it's strange to force the same port for internal and external access, but if the additional complexity is too high we can leave it like this for now

cluster/grpc_rpc_client_test.go
```diff
@@ -257,38 +257,107 @@ func TestSendPush(t *testing.T) {
...
+		server := &Server{
+			ID:   "someid",
+			Type: "sometype",
+			Metadata: map[string]string{
+				constants.GRPCHostKey: "localhost",
+				constants.GRPCPortKey: fmt.Sprintf("%d", port),
+			},
+			Frontend: false,
+		}
+		gs, err := NewGRPCServer(conf, server, []metrics.Reporter{})
```
golangcibot: ineffectual assignment to `err`

cluster/grpc_rpc_client.go
```diff
@@ -241,3 +249,33 @@ func (gs *GRPCClient) BeforeShutdown() {}
...
+		logger.Log.Warnf("server %s has no region nor external host specified in metadata, using internal host", sv.ID)
+		return internalHost, constants.GRPCPortKey
+	}
+
+	if gs.infoRetriever.Region() == serverRegion {
+		logger.Log.Warnf("server %s is in same region, using internal host", sv.ID)
+		return internalHost, constants.GRPCPortKey
+	}
+
+	logger.Log.Warnf("server %s is in other region, using external host", sv.ID)
```
andrehp: Should this be a warning level?

cluster/grpc_rpc_client.go
```diff
@@ -241,3 +249,33 @@ func (gs *GRPCClient) BeforeShutdown() {}
...
+
+func (gs *GRPCClient) getServerHost(sv *Server) (host, portKey string) {
+	var (
+		serverRegion, hasRegion   = sv.Metadata[constants.RegionKey]
+		externalHost, hasExternal = sv.Metadata[constants.GRPCExternalHostKey]
+		internalHost, _           = sv.Metadata[constants.GRPCHostKey]
+	)
+
+	hasRegion = hasRegion && serverRegion != ""
+	hasExternal = hasExternal && externalHost != ""
```
andrehp: What happens if the server doesn't specify external port but specifies external host?

Henrod: It'll fail here https://github.com/topfreegames/pitaya/blob/9bede6d29051472d79eade0f4a98076b12374fcc/cluster/grpc_rpc_client.go#L216

Henrod: It'll fail here: https://github.com/topfreegames/pitaya/blob/9bede6d29051472d79eade0f4a98076b12374fcc/cluster/grpc_rpc_client.go#L216

cluster/grpc_rpc_client.go
```diff
@@ -272,10 +272,10 @@ func (gs *GRPCClient) getServerHost(sv *Server) (host, portKey string) {
 	}
 
 	if gs.infoRetriever.Region() == serverRegion {
-		logger.Log.Warnf("server %s is in same region, using internal host", sv.ID)
+		logger.Log.Infof("server %s is in same region, using internal host", sv.ID)
```
andrehp: Don't you think that debug level would be better?

Henrod: Imagine if all connections are failing and we are struggling to debug (in a moment we can't put log level to debug). This will help a lot. Besides, it is only logged when a new server is up and all the others connect to it.

andrehp: Ok

andrehp:
Good job!

PR #60, opened 2018-09-25
Add descriptors

gabrielerzinger:
Add a function that returns a protobuf descriptor based on its message name or protobuf filename. This is used by pitaya admin in order to create RPC without prior knowledge about the `protos` that a game uses on its RPC as `input` or `output`.

Descriptors allow us to learn at runtime what fields a message contains and what the types of those fields are. They are registered at compile time, using its source file name `<abc>.proto` and can be retrieved by the same, or by calling the method `Descriptor` on a message.

Its in the same package as the docgenerator because they are related tools - both of them allow other services or APIs (such as pitaya admin) to manage information about pitaya servers. 

The game can use `pitaya.Descriptor(name)` to get the descriptor of a given protobuf message and expose it in a route for pitaya admin.

### DIFF
--- a/docgenerator/descriptors_test.go
+++ b/docgenerator/descriptors_test.go
@@ -1,0 +1,35 @@
+package docgenerator
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/topfreegames/pitaya/constants"
+	_ "github.com/topfreegames/pitaya/protos"
+)
+
+func TestProtoDescriptors(t *testing.T) {
+	t.Parallel()
+	tables := []struct {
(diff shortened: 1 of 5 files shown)


docgenerator/descriptors_test.go
```diff
@@ -0,0 +1,35 @@
+package docgenerator
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/topfreegames/pitaya/constants"
+	_ "github.com/topfreegames/pitaya/protos"
```
Henrod: Why did you import with underscore?

PR #66, opened 2018-10-31
Add pitaya admin to resources

gabrielerzinger:
I've added pitaya admin to README resources. I've also removed `Deployment` since I think it doesnt fit into a framework rep


andrehp:
GJ

PR #67, opened 2018-11-05
Feature/custom metrics

Henrod:
Possibility to add custom metrics on starting application. The danger in this is that a wrong config will cause the app to crash because the prometheus lib crashes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,9 @@ run-cluster-worker-example-metagame:
 run-cluster-worker-example-worker:
 	@cd examples/demo/worker && go run main.go --type worker --frontend=false
 
+run-custom-metrics-example:
+	@cd examples/demo/custom_metrics && go run main.go --port 3250
+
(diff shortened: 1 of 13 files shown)


andrehp:
Could you describe a little better how to use the custom metrics? The configuration values are not very clear for me

Henrod:
I don't know if doc is the better choice for this, maybe add an example?

andrehp:
Yeah, I guess an example would already help

Henrod:
@andrehp I made an example

andrehp:
Nice, could you just add a small paragraph to the docs mentioning custom metrics?

https://github.com/topfreegames/pitaya/blob/master/docs/features.md#monitoring

Henrod:
@andrehp do you think this is enough?

andrehp:
Yeah, LGTM

I'm gonna merge this, good job!

PR #69, opened 2018-11-12
Subgroup ETCD

rodrigolck:
**Description**

Added the feature of Subgroup from ETCD, which will enable the users to be inside a groups subgroup. All the methods that group has, so will the subgroup.

**How to Test**

The tests show exactly how to use it. But it is almost the same as the main group methods, with the subgroupName variable as a plus.


groups/etcd_group_service.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// GroupLeave removes specified UID from group
+func (c *EtcdGroupService) GroupLeave(ctx context.Context, groupName, uid string) error {
+	subgroups, err := c.PlayerSubgroups(ctx, groupName, uid)
+	if err != nil {
+		return err
+	}
+	for _, subgroup := range subgroups {
+		c.SubgroupLeave(ctx, groupName, subgroup, uid)
```
golangcibot: Error return value of `c.SubgroupLeave` is not checked

group.go
```diff
@@ -21,202 +21,182 @@
...
-func NewGroup(n string) *Group {
-	return &Group{
-		status:   groupStatusWorking,
-		name:     n,
-		sessions: make(map[string]*session.Session),
+// InitGroups should be called ONCE in the beginning of the application, if it is intended to use groups functionality
+func InitGroups(config *config.Config, clientOrNil *clientv3.Client) {
+	gsi, err := groups.NewEtcdGroupService(config, clientOrNil)
+	if err != nil {
+		logger.Log.Fatalf("error initializing groupServiceInstance in groups: %s", err.Error())
```
Henrod: you must return the error here

group.go
```diff
@@ -21,202 +21,182 @@
...
-}
-
-// NewGroup returns a new group instance
-func NewGroup(n string) *Group {
-	return &Group{
-		status:   groupStatusWorking,
-		name:     n,
-		sessions: make(map[string]*session.Session),
+// InitGroups should be called ONCE in the beginning of the application, if it is intended to use groups functionality
+func InitGroups(config *config.Config, clientOrNil *clientv3.Client) {
```
Henrod: Here you should receive `groupServiceOrNil` and create a EtcdGroupService if nil

groups/etcd_group_service.go
rodrigolck: Donne

PR #71, opened 2018-11-20
ETCD Group

rodrigolck:
**Description**

Added the possibility of creating groups in ETCD, which allows TTL. Also, made groups in memory have TTL to be used.

**How to Test**

The chat example is a good way of testing both implementations. Just need to change the initialisation of the service type inside the main.go


group_test.go
```diff
@@ -116,51 +171,69 @@ func TestLeave(t *testing.T) {
...
+			err = GroupAddMember(ctx, "testRemoveAll", s.UID())
+			assert.NoError(t, err)
+			err = GroupRemoveAll(ctx, "testRemoveAll")
+			assert.NoError(t, err)
+
+			res, err := GroupContainsMember(ctx, "testRemoveAllSufix", table.UID)
 			assert.NoError(t, err)
-			assert.False(t, g.Contains(table.UID))
+			assert.True(t, res)
+			res, err = GroupContainsMember(ctx, "testRemoveAll", table.UID)
```
golangcibot: ineffectual assignment to `res`

app_test.go
```diff
@@ -77,14 +79,31 @@ func setup() {
 	initApp()
 	Configure(true, "testtype", Cluster, map[string]string{}, viper.New())
 
-	etcdSD, _ := cluster.NewEtcdServiceDiscovery(app.config, app.server, app.dieChan)
+	etcdSD, err := cluster.NewEtcdServiceDiscovery(app.config, app.server, app.dieChan)
+	if err != nil {
+		panic(err)
```
andrehp: Panic in tests? =(

app_test.go
```diff
@@ -77,14 +79,31 @@ func setup() {
...
+	}
 	typeOfetcdSD = reflect.TypeOf(etcdSD)
 
-	natsRPCServer, _ := cluster.NewNatsRPCServer(app.config, app.server, nil, app.dieChan)
+	natsRPCServer, err := cluster.NewNatsRPCServer(app.config, app.server, nil, app.dieChan)
+	if err != nil {
+		panic(err)
```
andrehp: Panic in tests? =(

app_test.go
```diff
@@ -77,14 +79,31 @@ func setup() {
...
+	}
 	typeOfNatsRPCServer = reflect.TypeOf(natsRPCServer)
 
-	natsRPCClient, _ := cluster.NewNatsRPCClient(app.config, app.server, nil, app.dieChan)
+	natsRPCClient, err := cluster.NewNatsRPCClient(app.config, app.server, nil, app.dieChan)
+	if err != nil {
+		panic(err)
```
andrehp: Panic in tests? =(

app_test.go
```diff
@@ -77,14 +79,31 @@ func setup() {
...
+	}
 	typeOfNatsRPCClient = reflect.TypeOf(natsRPCClient)
+
+	c := integration.NewClusterV3(nil, &integration.ClusterConfig{Size: 1})
+	cli := c.RandClient()
+	gsi, err := groups.NewEtcdGroupService(app.config, cli)
+	if err != nil {
+		panic(err)
```
andrehp: Panic in tests? =(

constants/errors.go
```diff
@@ -36,11 +36,16 @@ var (
 	ErrCloseClosedSession             = errors.New("close closed session")
 	ErrClosedGroup                    = errors.New("group closed")
 	ErrEtcdGrantLeaseTimeout          = errors.New("timed out waiting for etcd lease grant")
+	ErrEtcdLeaseNotFound              = errors.New("etcd lease not found in group")
```
andrehp: Try to leave the errors in alphabetical order, and take the opportunity to fix the comment at the top of the file

constants/errors.go
```diff
@@ -36,11 +36,16 @@ var (
...
 	ErrFrontSessionCantPushToFront    = errors.New("frontend session can't push to front")
 	ErrIllegalUID                     = errors.New("illegal uid")
 	ErrKickingUsers                   = errors.New("failed to kick users, check array with failed uids")
 	ErrPushingToUsers                 = errors.New("failed to push message to users, check array with failed uids")
+	ErrMemoryTTLNotFound              = errors.New("memory group TTL not found")
 	ErrMemberNotFound                 = errors.New("member not found in the group")
+	ErrMemberDuplication              = errors.New("member already exists in group")
+	ErrGroupNotFound                  = errors.New("group not found")
+	ErrGroupDuplication               = errors.New("group already exists")
```
andrehp: I think that ErrGroupAlreadyExists would be better than ErrGroupDuplication

docs/configuration.rst
```diff
@@ -371,3 +371,32 @@ These configurations control if the default pipelines should be enabled or not
...
+    - string
+    - Comma separated list of etcd endpoints to be used by the groups etcd service 
+  * - pitaya.groups.etcd.prefix
+    - pitaya/
+    - string 
+    - Prefix used for every group key in etcd
+  * - pitaya.groups.etcd.dialtimeout
+    - 5s
+    - time.Time
+    - Timeout to establish the etcd connection
```
andrehp: Timeout to establish the etcd group connection

groups/etcd_group_service.go
```diff
@@ -0,0 +1,215 @@
...
+		_, err = clientInstance.Put(ctx, groupKey(groupName), string(jsonPayload))
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GroupCreate returns all member's UID and payload in current group
```
andrehp: I think this comment is wrong

groups/memory_group_service.go
```diff
@@ -0,0 +1,185 @@
...
+
+	mg, ok := memoryGroups[groupName]
+	if !ok {
+		return nil, constants.ErrGroupNotFound
+	}
+
+	return mg, nil
+}
+
+// GroupCreate returns all member's UID and payload in current group
```
andrehp: Comment is wrong

groups/memory_group_service.go
```diff
@@ -0,0 +1,185 @@
...
+	LastRefresh int64
+	TTL         int64
+}
+
+// NewMemoryGroupService returns a new group instance
+func NewMemoryGroupService() *MemoryGroupService {
+	memoryOnce.Do(func() {
+		memoryGroups = make(map[string]*MemoryGroup)
+		go func() {
+			for now := range time.Tick(time.Second) {
```
andrehp: Wouldn't it be better for this interval to be in a config?

groups/group_service.go
```diff
@@ -0,0 +1,30 @@
+package groups
+
+import (
+	"context"
+	"time"
+)
+
+type (
+	// GroupService has ranking methods
+	GroupService interface {
```
felipejfc: shouldn't we have a GroupDelete method?

groups/etcd_group_service.go
```diff
@@ -0,0 +1,215 @@
...
+		return constants.ErrMemberAlreadyExists
+	}
+
+	payload.Uids = append(payload.Uids, uid)
+	return putGroupPayload(ctx, groupName, payload)
+}
+
+// GroupRemoveMember removes specified UID from group
+func (c *EtcdGroupService) GroupRemoveMember(ctx context.Context, groupName, uid string) error {
+	payload, err := getGroupPayload(ctx, groupName)
```
felipejfc: I guess there's a running condition that can happen here?

GroupAddMember and GroupRemoveMember are not atomic, both of them are doing the following:

read
process
write

If another operation begins while one is in the process phase we will have an undefined behaviour.

groups/memory_group_service.go
```diff
@@ -0,0 +1,186 @@
...
+		return constants.ErrGroupAlreadyExists
+	}
+
+	memoryGroups[groupName] = &MemoryGroup{LastRefresh: time.Now().UnixNano(), TTL: ttlTime.Nanoseconds()}
+	return nil
+}
+
+// GroupMembers returns all member's UID in given group
+func (c *MemoryGroupService) GroupMembers(ctx context.Context, groupName string) ([]string, error) {
+	mg, err := getMemoryGroup(groupName)
```
felipejfc: shouldn't you use the lock here?
also it seems that it would be more optimal if you have a group-specific mutex for changes like add and remove members and another one that is global for groupcreate and delete...

groups/memory_group_service.go
```diff
@@ -0,0 +1,186 @@
...
+		return constants.ErrGroupNotFound
+	}
+
+	delete(memoryGroups, groupName)
+	return nil
+}
+
+// GroupCountMembers get current member amount in group
+func (c *MemoryGroupService) GroupCountMembers(ctx context.Context, groupName string) (int, error) {
+	mg, err := getMemoryGroup(groupName)
```
felipejfc: shouldn't the lock be used here too?

groups/group_service.go
rodrigolck: We have GroupRemoveAll, which removes everything, including the group itself. Should I change the behaviour to GroupRemoveAll remove all the uids, but not delete and GroupDelete delete it everything?

felipejfc: Yes, I guess this would be more clear

groups/memory_group_service.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// GroupMembers returns all member's UID in given group
+func (c *MemoryGroupService) GroupMembers(ctx context.Context, groupName string) ([]string, error) {
+	mg, err := getMemoryGroup(groupName)
+	if err != nil {
+		return nil, err
+	}
+
+	return mg.Uids, nil
```
andrehp: Here you return the array of ids, which is an array, any function that uses the return value of this function is able to modify the group elements, you should ideally return a copy of the array

groups/memory_group_service.go
rodrigolck: Changed the implementations, to the point that every operation will be atomic. About the implementation of global lock and local lock, it would introduce a large complexity into the code, which can also cause more harm than good. Another solution would be to use sync.Map, but it would also be a global lock...

groups/memory_group_service.go
rodrigolck: Thanks for the advice! Changed the implementation and copied the slice values to another one!

groups/etcd_group_service.go
rodrigolck: Understood the problem! Changed the implementation to be everything atomic, so that these types of errors don't occur

groups/group_service.go
rodrigolck: Done!

groups/etcd_group_service.go
```diff
@@ -0,0 +1,254 @@
...
+	cli.KV = namespace.NewKV(cli.KV, config.GetString("pitaya.groups.etcd.prefix"))
+	return cli, nil
+}
+
+func groupKey(groupName string) string {
+	return fmt.Sprintf("groups/%s", groupName)
+}
+
+func memberKey(groupName, uid string) string {
+	return groupKey(groupName) + fmt.Sprintf("/uids/%s", uid)
```
andrehp: Don't concatenate strings like this

groups/etcd_group_service.go
```diff
@@ -0,0 +1,254 @@
...
+
+	getRes := etcdRes.Responses[0].GetResponseRange()
+	members := make([]string, getRes.GetCount())
+	for i, kv := range getRes.GetKvs() {
+		members[i] = string(kv.Key)[len(prefix):]
+	}
+	return members, nil
+}
+
+// GroupContainsMember check whether a UID is contained in current group or not
```
andrehp: s/check/checks

groups/group_service.go
```diff
@@ -0,0 +1,31 @@
...
+		GroupCreate(ctx context.Context, groupName string) error
+		GroupCreateWithTTL(ctx context.Context, groupName string, ttlTime time.Duration) error
+		GroupDelete(ctx context.Context, groupName string) error
+		GroupMembers(ctx context.Context, groupName string) ([]string, error)
+		GroupContainsMember(ctx context.Context, groupName, uid string) (bool, error)
+		GroupAddMember(ctx context.Context, groupName, uid string) error
+		GroupRemoveMember(ctx context.Context, groupName, uid string) error
+		GroupRemoveAll(ctx context.Context, groupName string) error
+		GroupCountMembers(ctx context.Context, groupName string) (int, error)
+		GroupRenewTTL(ctx context.Context, groupName string) error
```
andrehp: What do you think about sorting the method names alphabetically?

group.go
```diff
@@ -21,202 +21,98 @@
...
+var groupServiceInstance groups.GroupService
 
-// NewGroup returns a new group instance
-func NewGroup(n string) *Group {
-	return &Group{
-		status:   groupStatusWorking,
-		name:     n,
-		sessions: make(map[string]*session.Session),
-	}
+// InitGroups should be called once at the beginning of the application to setup the service type that will manage the groups
```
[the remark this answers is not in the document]
rodrigolck: Okie Dokie!

groups/etcd_group_service.go
```diff
@@ -0,0 +1,284 @@
...
+			If(clientv3.Compare(clientv3.CreateRevision(groupKey(groupName)), "=", 0)).
+			Then(clientv3.OpPut(groupKey(groupName), "", clientv3.WithLease(leaseID))).
+			Commit()
+	} else {
+		etcdRes, err = clientInstance.Txn(ctxT).
+			If(clientv3.Compare(clientv3.CreateRevision(groupKey(groupName)), "=", 0)).
+			Then(clientv3.OpPut(groupKey(groupName), "")).
+			Commit()
+	}
+	cancel()
```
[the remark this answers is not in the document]
rodrigolck: This is an interesting question. If I use defer, it will only be called at the end of the whole function which it has been called and I only want the transaction operation to be with timeout. That's why I inserted it right after the method call, to be certain that it was called exactly after. Also, the doc says that the timeout should be called asap. Should I use defer?

group.go
```diff
@@ -21,202 +21,98 @@
...
 }
 
-// Contains check whether a UID is contained in current group or not
-func (c *Group) Contains(uid string) bool {
-	_, err := c.Member(uid)
-	return err == nil
+// GroupContainsMember checks whether an UID is contained in group or not
+func GroupContainsMember(ctx context.Context, groupName, uid string) (bool, error) {
+	if uid == "" {
+		return false, constants.ErrNoUIDBind
```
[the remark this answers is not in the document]
rodrigolck: What do you think about: constants.ErrEmptyUID ?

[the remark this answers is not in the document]
cscatolini: better :)

groups/memory_group_service.go
```diff
@@ -0,0 +1,201 @@
...
+	Uids        []string
+	LastRefresh int64
+	TTL         int64
+}
+
+// NewMemoryGroupService returns a new group instance
+func NewMemoryGroupService(conf *config.Config) *MemoryGroupService {
+	memoryOnce.Do(func() {
+		memoryGroups = make(map[string]*MemoryGroup)
+		go func() {
```
felipejfc: create a new function for that, it's a little ugly inside the constructor

rodrigolck: Okie Dokie!

groups/memory_group_service.go
```diff
@@ -0,0 +1,205 @@
...
+	Uids        []string
+	LastRefresh int64
+	TTL         int64
+}
+
+// NewMemoryGroupService returns a new group instance
+func NewMemoryGroupService(conf *config.Config) *MemoryGroupService {
+	memoryOnce.Do(func() {
+		memoryGroups = make(map[string]*MemoryGroup)
+		groupTTLCleanup(conf)
```
felipejfc: it's better to do go groupTTLCleanup(conf)

rodrigolck: Okie Dokie!

cscatolini:
before we can merge this, can you take a look why the tests are failing? @rodrigolck

rodrigolck:
Took a look and saw that it is the e2e tests that are failing. Need to figure out why the timeouts are happening

cscatolini:
@rodrigolck don't worry about the flaky tests. We're good to merge this.

PR #72, opened 2018-11-22
Message package be public

rodrigolck:
**Description**

Found out that the message package is private, which makes lots of other projects that use it unable to  access it properly. It creates difficulties such as knowing the message type, how to handle it and so on. One example is the pitaya bot project.

**How to Test it**

See if everything continues working correctly and see if it is viable to merge what I am proposing. If not, why.


andrehp:
I believe this change makes sense, either expose this package or export it to a different project so that other projects can use the information as well

PR #75, opened 2018-12-14
Add a new pitaya client with protocol buffers support.

capellaresumo:


client/client.go
```diff
@@ -90,6 +90,14 @@ type Client struct {
 	messageEncoder  message.Encoder
 }
 
+func (c *Client) MsgChannel() chan *message.Message {
```
andrehp: Please add the comment description to this and the other functions

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+	protobuf "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/dynamic"
+	"github.com/sirupsen/logrus"
+	"github.com/topfreegames/pitaya/conn/message"
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/protos"
+)
+
+type ClientInterface interface {
```
andrehp: Move this interface definition to a separate file

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+
+	err := pc.buildProtosFromDescriptor(descriptorArray)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func new(docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
```
andrehp: A better name would be newProto

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+
+func NewWithDescriptor(descriptorsRoute string, docsRoute string, docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
+	newclient := new(docslogLevel, requestTimeout...)
+	newclient.docsRoute = docsRoute
+	newclient.descriptorsRoute = descriptorsRoute
+	return newclient
+}
+
+// Load commands information form the server. Names is a list of protos names.
+func (pc *ProtoClient) LoadServoInfo(addr string) error {
```
andrehp: Servo?

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+}
+
+func NewWithDescriptor(descriptorsRoute string, docsRoute string, docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
+	newclient := new(docslogLevel, requestTimeout...)
+	newclient.docsRoute = docsRoute
+	newclient.descriptorsRoute = descriptorsRoute
+	return newclient
+}
+
+// Load commands information form the server. Names is a list of protos names.
```
andrehp: s/form/from

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func new(docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
+	return &ProtoClient{
+		Client:           *New(docslogLevel, requestTimeout...),
```
andrehp: New already returns a pointer

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+			if strings.Contains(k, "proto") {
+				outputName = strings.Replace(k, "*", "", 1)
+			}
+		}
+	}
+
+	return inputName, outputName, nil
+}
+
+// get recursivily all protos needed in a Unmarshal json
```
andrehp: s/recursivily/recursively

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+
+		var command Command
+		command.input = in
+		command.output = out
+
+		pc.info.Commands[k] = &command
+		if pc.descriptorsRoute == "" && in == "protos.ProtoName" && out == "protos.ProtoDescriptor" {
+			pc.descriptorsRoute = k
+		}
+		// c.Println(k);
```
andrehp: Remove this line

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+		if err != nil {
+			return err
+		}
+
+		var command Command
+		command.input = in
+		command.output = out
+
+		pc.info.Commands[k] = &command
+		// c.Println(k);
```
andrehp: remove this line

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+
+	for key := range keysSet {
+		protname := &protos.ProtoName{
+			Name: key,
+		}
+		data, err := proto.Marshal(protname)
+		if err != nil {
+			return err
+		}
+		_, err = pc.SendRequest(pc.descriptorsRoute, data)
```
andrehp: Do you need to send a request for each proto?

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+	for {
+		select {
+		case response := <-pc.Client.IncomingMsgChan:
+
+			inputMsg := pc.expectedInput
+
+			msg, ok := pc.info.Commands[response.Route]
+			if ok {
+				inputMsg = msg.outputMsg
+				// fmt.Println(msg)
```
andrehp: Remove this line

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+func (pc *ProtoClient) ConnectTo(addr string) error {
+	err := pc.Client.ConnectTo(addr)
+	if err != nil {
+		return err
+	}
+	go pc.waitForData()
+	return nil
+}
+
+// Export sup[orted commands information
```
andrehp: s/sup[orted/supported

client/protoclient.go
andrehp: And when should this function be called?

client/protoclient.go
```diff
@@ -0,0 +1,491 @@
...
+// Disconnect the client
+func (pc *ProtoClient) Disconnect() {
+	pc.Client.Disconnect()
+	if pc.ready {
+		pc.closeChan <- true
+	}
+}
+
+func (pc *ProtoClient) waitForData() {
+	if !pc.ready {
```
andrehp: Wouldn't it be better to initialize the client if it isn't yet instead of silently returning?

client/protoclient.go
capellaresumo: ProtoClient extends the Client struct. Here I get the value of the pointer returned by New.

client/protoclient.go
```diff
@@ -0,0 +1,523 @@
...
+func newProto(docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
+	return &ProtoClient{
+		Client:           *New(docslogLevel, requestTimeout...),
+		descriptorsNames: make(map[string]bool),
+		info: ProtoBufferInfo{
+			Commands: make(map[string]*Command),
+		},
+		docsRoute:        "",
+		descriptorsRoute: "",
+		IncomingMsgChan:  make(chan *message.Message, 3),
```
andrehp: Any special reason for the value 3?

client/protoclient.go
```diff
@@ -0,0 +1,523 @@
...
+// NewWithDescriptor returns a new protoclient with the descriptors route and
+// auto documentation route.
+func NewWithDescriptor(descriptorsRoute string, docsRoute string, docslogLevel logrus.Level, requestTimeout ...time.Duration) *ProtoClient {
+	newclient := newProto(docslogLevel, requestTimeout...)
+	newclient.docsRoute = docsRoute
+	newclient.descriptorsRoute = descriptorsRoute
+	return newclient
+}
+
+// LoadServerInfo load commands information from the server. Addr is ther
```
andrehp: s/ther/the

client/protoclient.go
capellaresumo: The main reason is to create a small buffer. I will change it to 10, to be equal to the client/client.go

PR #77, opened 2019-01-08
Changing tracing to work with elasticapm

andrehp:


tracing/span.go
```diff
@@ -21,15 +21,29 @@
...
 
+func castValueToCarrier(val interface{}) (opentracing.TextMapCarrier, error) {
+	if v, ok := val.(opentracing.TextMapCarrier); ok {
+		return v, nil
+	}
+	if m, ok := val.(map[string]interface{}); ok {
+		carrier := map[string]string{}
+		for k, v := range m {
+			if s, ok := v.(string); ok {
+				carrier[k] = s
```
cscatolini: If some of the keys cannot be converted to a string it will be ignored? Should we log something?

andrehp: That's a good point, avoids silent errors, I'm going to add a log here

PR #87, opened 2019-02-13
Fix error handler - protoclient

capellaresumo:
Before this commit when one request returns one error, the
error was decode like the expected message. This commit
solves this problem.


felipejfc:
awesome! thanks!

PR #90, opened 2019-02-14
Make the correct use of the protoreflect

capellaresumo:
This commit fixes the problem of protobuffer client when
running in multiple goroutines.

github.com/jhump/protoreflect has the desc.MessageDescriptor
and dynamic.Message. Before this commit, the dynamic.Message
struct was associated with the expected input. After it, the
message will be associated whit one desc.MessageDescriptor
and when some data is received, the message struct is
created. That is, each input package will have only one
dynamic.Message.

### DIFF
--- a/client/protoclient.go
+++ b/client/protoclient.go
@@ -42,10 +42,10 @@ import (
 // Command struct. Save the input and output type and proto descriptor for each
 // one.
 type Command struct {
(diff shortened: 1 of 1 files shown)


felipejfc:
awesome! thanks @capellaresumo

PR #92, opened 2019-02-26
Report code on metrics when error

Henrod:
Send code on metrics report when error. This helps filtering errors by using the tag "code". 

Code is reported empty in case of not error. I'm not sure if this is the best thing to do but pitaya doesn't enforce code on responses, only on errors. 

Since the tests will come on next commits, this is a WIP.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -23,7 +23,7 @@ package agent
 import (
 	"context"
 	gojson "encoding/json"
-	"errors"
+	e "errors"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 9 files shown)


agent/agent.go
```diff
@@ -394,15 +395,23 @@ func (a *Agent) write() {
...
 			}
 			var e error
 			tracing.FinishSpan(data.ctx, e)
 			if data.typ == message.Response {
-				metrics.ReportTimingFromCtx(data.ctx, a.metricsReporters, handlerType, m.Err)
+				var rErr error
+
+				if m.Err {
+					rErr = &errors.Error{}
+					// TODO: should I ignore this error?
```
Henrod: I intend to remove this comment after some discussion.

andrehp: What's the result if m.Data is not a valid error?

Henrod: That would be a mismatch between data.payload and data.err. Is it possible? But I could check if code is not empty to continue.

andrehp: Maybe set a default value just to be safe?

PR #96, opened 2019-03-19
Fix/service discovery

### LINKED ISSUE #95: Concurrent access to servers list leads to panic
Hi,

A concurrency bug can lead to server crashes in very specific circumstances when acessing and iterating over the map of servers by type returned.

The map returned by `GetServersByType`, from etcd's service discovery implementation is the same that is updated by the service discovery routine. This leads to possible concurrent read and write if the service discovery routine updates the map at the same time that another function is accessing the map, every request received by a front end server calls `GetServersByType`, so this is probably the most common user of the method.

Returning the map which is modified by the service can also lead to misuse of this structure and inconsistencies in the service discovery data.

andrehp:
Fixes #95

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -232,15 +232,21 @@ func (sd *etcdServiceDiscovery) notifyListeners(act Action, sv *Server) {
 	}
 }
 
+func (sd *etcdServiceDiscovery) writeLockScope(f func()) {
+	sd.mapByTypeLock.Lock()
+	defer sd.mapByTypeLock.Unlock()
+	f()
+}
+
 func (sd *etcdServiceDiscovery) deleteServer(serverID string) {
 	if actual, ok := sd.serverMapByID.Load(serverID); ok {
 		sv := actual.(*Server)
 		sd.serverMapByID.Delete(sv.ID)
-		sd.mapByTypeLock.Lock()
-		if svMap, ok := sd.serverMapByType[sv.Type]; ok {
-			delete(svMap, sv.ID)
-		}
-		sd.mapByTypeLock.Unlock()
+		sd.writeLockScope(func() {
+			if svMap, ok := sd.serverMapByType[sv.Type]; ok {
+				delete(svMap, sv.ID)
+			}
+		})
 		sd.notifyListeners(DEL, sv)
 	}
 }
@@ -386,10 +392,10 @@ func parseServer(value []byte) (*Server, error) {
 
 func (sd *etcdServiceDiscovery) printServers() {
 	sd.mapByTypeLock.RLock()
+	defer sd.mapByTypeLock.RUnlock()
 	for k, v := range sd.serverMapByType {
 		logger.Log.Debugf("type: %s, servers: %+v", k, v)
 	}
-	sd.mapByTypeLock.RUnlock()
 }
 
 // SyncServers gets all servers from etcd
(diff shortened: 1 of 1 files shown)


cluster/etcd_service_discovery.go
```diff
@@ -74,13 +76,15 @@ func NewEtcdServiceDiscovery(
...
-		running:    false,
-		server:     server,
-		listeners:  make([]SDListener, 0),
-		stopChan:   make(chan bool),
-		appDieChan: appDieChan,
-		cli:        client,
+		config:          config,
+		running:         false,
+		server:          server,
+		serverMapByType: make(map[string]map[string]*Server, 0),
```
golangcibot: S1019: should use make(map[string]map[string]*Server) instead (from `gosimple`)

cluster/etcd_service_discovery.go
```diff
@@ -230,10 +236,11 @@ func (sd *etcdServiceDiscovery) deleteServer(serverID string) {
 	if actual, ok := sd.serverMapByID.Load(serverID); ok {
 		sv := actual.(*Server)
 		sd.serverMapByID.Delete(sv.ID)
-		if svMap, ok := sd.serverMapByType.Load(sv.Type); ok {
-			sm := svMap.(map[string]*Server)
-			delete(sm, sv.ID)
+		sd.mapByTypeLock.Lock()
```
cscatolini: Isn't it better to `defer sd.mapByTypeLock.Unlock()` right after locking?

cluster/etcd_service_discovery.go
```diff
@@ -267,10 +274,15 @@ func (sd *etcdServiceDiscovery) getServerFromEtcd(serverType, serverID string) (
 
 // GetServersByType returns a slice with all the servers of a certain type
 func (sd *etcdServiceDiscovery) GetServersByType(serverType string) (map[string]*Server, error) {
-	if m, ok := sd.serverMapByType.Load(serverType); ok {
-		sm := m.(map[string]*Server)
-		if len(sm) > 0 {
-			return sm, nil
+	sd.mapByTypeLock.RLock()
+	defer sd.mapByTypeLock.RUnlock()
+	if m, ok := sd.serverMapByType[serverType]; ok {
```
cscatolini: I don't like the nested if. Do you know if this works?

```
if m, ok := sd.serverMapByType[serverType]; ok && len(m) > 0 {
  ...
}
```

cluster/etcd_service_discovery.go
```diff
@@ -267,10 +274,15 @@ func (sd *etcdServiceDiscovery) getServerFromEtcd(serverType, serverID string) (
...
+		if len(m) > 0 {
+			ret := make(map[string]*Server, len(sd.serverMapByType))
+			for k, v := range sd.serverMapByType[serverType] {
```
cscatolini: It is not obvious why your are iterating over a map to build the exact same map. I have the context by what we've talked and I had to think for a while before remembering it. I'd leave a comment here explaining why this is necessary.

cluster/etcd_service_discovery.go
```diff
@@ -374,10 +384,11 @@ func parseServer(value []byte) (*Server, error) {
 }
 
 func (sd *etcdServiceDiscovery) printServers() {
-	sd.serverMapByType.Range(func(k, v interface{}) bool {
-		logger.Log.Debugf("type: %s, servers: %s", k, v)
-		return true
-	})
+	sd.mapByTypeLock.RLock()
```
cscatolini: Same here. Why not use `defer`?

cluster/etcd_service_discovery.go
```diff
@@ -454,12 +466,14 @@ func (sd *etcdServiceDiscovery) revoke() error {
 
 func (sd *etcdServiceDiscovery) addServer(sv *Server) {
 	if _, loaded := sd.serverMapByID.LoadOrStore(sv.ID, sv); !loaded {
-		mapSvByType, ok := sd.serverMapByType.Load(sv.Type)
+		sd.mapByTypeLock.Lock()
```
cscatolini: Defer?

cluster/etcd_service_discovery.go
andrehp: Given that there is only one flow deferring is slower and would be executed only after notifying the listeners

cluster/etcd_service_discovery.go
andrehp: Yeah, I agree this looks better, I'm going to update it

cluster/etcd_service_discovery.go
andrehp: Good idea

cluster/etcd_service_discovery.go
andrehp: Defers are only useful when there multiple flows, in simple cases like this calling the function directly is a little more efficient

cluster/etcd_service_discovery.go
andrehp: Same as above, using defer would keep the lock longer than needed

cscatolini: Same as my reply in the first comment about defers.

cluster/etcd_service_discovery.go
andrehp: Ok, I created a wrapper for the lock, what do you think?

cscatolini:
Great job 😍

PR #102, opened 2019-05-22
rate limits on router

Henrod:
Use WithRateLimiting function as RoutingFunc decorator
- It's optional, without the decorator no rate limit is not applied
- On disastrous situations in production, WithRateLimiting can be forcibly disabled by config
- The number of blocked requests can be monitored by the code `PIT-429` on `response_time_ns` metric
- RateLimiting is saved on player's session, so no bind is necessary to control rate


router/rate_limiting.go
```diff
@@ -0,0 +1,109 @@
...
+			rl  *RateLimiting
+			s   = sessionVal.(*session.Session)
+			now = time.Now()
+		)
+
+		if s.HasKey(constants.RateLimitingKey) {
+			rl = s.Get(constants.RateLimitingKey).(*RateLimiting)
+		} else {
+			rl = NewRateLimiting(limit, interval)
+			s.Set(constants.RateLimitingKey, rl)
```
golangcibot: Error return value of `s.Set` is not checked (from `errcheck`)

router/rate_limiting_test.go
```diff
@@ -0,0 +1,190 @@
...
+		before func()
+		err    error
+	}{
+		"test_can_take_on_first_time": {
+			before: func() {},
+			err:    nil,
+		},
+		"test_can_take_missing_one_to_limit": {
+			before: func() {
+				r.Take(now)
```
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

router/rate_limiting_test.go
```diff
@@ -0,0 +1,190 @@
...
+				r.Take(now)
```
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

router/rate_limiting_test.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+			err: nil,
+		},
+		"test_can_take_when_oldest_request_expired": {
+			before: func() {
+				r.Take(now.Add(-2 * interval))
```
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

config/config.go
```diff
@@ -105,6 +105,8 @@ func (c *Config) fillDefaultValues() {
 		"pitaya.modules.bindingstorage.etcd.endpoints":     "localhost:2379",
 		"pitaya.modules.bindingstorage.etcd.leasettl":      "1h",
 		"pitaya.modules.bindingstorage.etcd.prefix":        "pitaya/",
+		"pitaya.router.ratelimiting.limit":                 20,
+		"pitaya.router.ratelimiting.interval":              "1s",
```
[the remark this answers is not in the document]
Henrod: if not set, the default is false for bool

Henrod:
> Why use a wrapper function and force the users to use a custom routing function instead of simply calling the RateLimiting function if rate limiting is enabled?

Do you think rate limiting should be enabled by default? But it's a good point that defaultRouting also should be able do use rate limiting, I will work on that.

router/rate_limiting_test.go
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

router/rate_limiting_test.go
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

router/rate_limiting_test.go
golangcibot: Error return value of `r.Take` is not checked (from `errcheck`)

cscatolini:
@Henrod I don't think the routing function is the correct place to handle this. We're trying to avoid badly behaved clients, correct? If we put this logic in the routing function we're only limiting calls that need to be redirected to other servers. 

IMO rate limiting should be enforced as close as possible to the client, i.e. in the frontendend server she is connected to. In http APIs rate limiting are usually done in a proxy that is reached before the API. I would move this logic to somewhere in `HandlerService`, as soon as we have the client information so we can limit by client.

Henrod:
> @Henrod I don't think the routing function is the correct place to handle this. We're trying to avoid badly behaved clients, correct? If we put this logic in the routing function we're only limiting calls that need to be redirected to other servers.
> 
> IMO rate limiting should be enforced as close as possible to the client, i.e. in the frontendend server she is connected to. In http APIs rate limiting are usually done in a proxy that is reached before the API. I would move this logic to somewhere in `HandlerService`, as soon as we have the client information so we can limit by client.

Good point!

Henrod:
@cscatolini @andrehp So, I am thinking about checking for rateLimiting [here](https://github.com/topfreegames/pitaya/blob/feature/rate-limiting/service/handler.go#L290). What do you think? Just before sending to localProcess and remoteProcess

Also, I am not sure where to put *RateLimiting. Another file in service directory? Or a module?

config/config.go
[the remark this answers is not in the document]
felipejfc: this file should contain all defaults for reference, so add it here and if you included any other config and it's default is not here, include it as well

docs/features.md
```diff
@@ -29,6 +29,9 @@ When a server instance receives a client message, it checks the target server ty
 
 By default the routing function chooses one instance of the target server type at random. Custom functions can be defined to change this behavior.
 
+#### Rate limiting
+By adding a routing function, it is possible to wrap it with a rate limiting logic by using WithRateLimiting function. The requests counts and management is done on player's session, therefore it happens even before session bind. The used algorithm is the Leaky Bucket.
```
felipejfc: include a link with a description of the leaky bucket algorithm

router/rate_limiting.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/topfreegames/pitaya/constants"
+	"github.com/topfreegames/pitaya/errors"
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/route"
+	"github.com/topfreegames/pitaya/session"
+)
+
+// WithRateLimiting wraps routingFunc and controls
+// player rate through session.
+func WithRateLimiting(c *config.Config, f RoutingFunc) RoutingFunc {
```
felipejfc: is it possible to use an implementation of a pitaya pipeline instead of a decorator function?

router/rate_limiting.go
```diff
@@ -0,0 +1,109 @@
...
+		logger.Log.Warn("rate limiting is forcibly disabled")
+	}
+
+	return func(
+		ctx context.Context,
+		r *route.Route,
+		payload []byte,
+		servers map[string]*cluster.Server,
+	) (*cluster.Server, error) {
+		if forceDisable {
```
felipejfc: this is an optional feature that one had intentionally included, is the point of having this config just to make it easier to turn it off without having to make a new build?

router/rate_limiting.go
```diff
@@ -0,0 +1,109 @@
...
+			return f(ctx, r, payload, servers)
+		}
+
+		sessionVal := ctx.Value(constants.SessionCtxKey)
+		if sessionVal == nil {
+			logger.Log.Warnf("session not found, cant rate limit, route: %s", r.String())
```
felipejfc: "can't"

also, when would this case happen (not having a session)?

router/rate_limiting.go
```diff
@@ -0,0 +1,109 @@
...
+
+	r.times.Init()
+
+	return r
+}
+
+// Take saves the now as time taken or returns an error if
+// in the limit of rate limiting
+func (r *RateLimiting) Take(now time.Time) error {
+	r.mutex.Lock()
```
felipejfc: do you think it's possible to remove this mutex?

cscatolini:
@felipejfc I also thought about pipelines but it doesn't fit the requirement of blocking the message as soon as possible. If the handler service is the one responsible for handling messages, I think it is okay to make it responsibe for blocking messages.

router/rate_limiting.go
Henrod: Yes. In case you notice that players are being blocked by rate limiting unintentionally and you want to disable it for now to test other values for limit and interval

router/rate_limiting.go
Henrod: Just for safety, so the application doesn't crash. But I will verify if this can't happen at all and, if so, I remove this check

felipejfc:
> @felipejfc I also thought about pipelines but it doesn't fit the requirement of blocking the message as soon as possible. If the handler service is the one responsible for handling messages, I think it is okay to make it responsibe for blocking messages.

What I meant is that, handlerService's only responsibility should be to receive messages from clients, decoding it and processing it. Today, apart from that it's also dealing with handshake packets, I'ld like to refactor that.

Henrod:
@felipejfc If filtering after routing, the logic would be at `removeService.remoteProcess` method and `handlerService.localProcess` method. That would work.

andrehp:
But then you would need to add it at two distinct places in the code instead of only one

felipejfc:
Discussing it with @Henrod I think we came up with a good solution, the idea is to decorate the acceptor with a concept of ConnectionWrappers that could be used for rate limiting or other stuff like regional ip blocking for example.
This way we keep the right responsibilities for each component and the code for the rate limiter will be explicitly "plugable" and easy to find and maintain.

acceptorwrapper/base_test.go
```diff
@@ -0,0 +1,52 @@
...
+			conns <- mockConn
+		}
+	}()
+
+	mockAcceptor.EXPECT().GetConnChan().Return(conns)
+	wrapper := &BaseWrapper{
+		Acceptor: mockAcceptor,
+		connChan: make(chan net.Conn),
+		wrapConn: func(c net.Conn) net.Conn {
+			c.Read([]byte{})
```
golangcibot: Error return value of `c.Read` is not checked (from `errcheck`)

acceptorwrapper/rate_limiting_conn_test.go
```diff
@@ -0,0 +1,150 @@
...
+		before func()
+		err    error
+	}{
+		"test_can_take_on_first_time": {
+			before: func() {},
+			err:    nil,
+		},
+		"test_can_take_missing_one_to_limit": {
+			before: func() {
+				r.take(now)
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

acceptorwrapper/rate_limiting_conn_test.go
```diff
@@ -0,0 +1,150 @@
...
+				r.take(now)
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

acceptorwrapper/rate_limiting_conn_test.go
```diff
@@ -0,0 +1,150 @@
...
+			},
+			err: nil,
+		},
+		"test_can_take_when_oldest_request_expired": {
+			before: func() {
+				r.take(now.Add(-2 * interval))
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

metrics/report.go
```diff
@@ -94,6 +94,14 @@ func ReportSysMetrics(reporters []Reporter, period time.Duration) {
 	}
 }
 
+// ReportExceededRateLimiting reports the number of requests made
+// after exceeded rate limiting in a connection
+func ReportExceededRateLimiting(reporters []Reporter) {
+	for _, r := range reporters {
+		r.ReportCount(ExceededRateLimiting, map[string]string{}, 1)
```
golangcibot: Error return value of `r.ReportCount` is not checked (from `errcheck`)

acceptorwrapper/rate_limiting_conn.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return r
+}
+
+func (r *RateLimitingConn) Read(b []byte) (n int, err error) {
+	if r.forceDisable {
+		return r.Conn.Read(b)
+	}
+
+	n, err = r.Conn.Read(b)
```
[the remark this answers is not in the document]
Henrod: But that's what happens now, [see](https://github.com/topfreegames/pitaya/blob/master/service/handler.go#L184)

Henrod:
> The tests are missing from this PR. An example would be nice as well.

Made the example on commit b6d10c3e96bf8d45fc543766db27a195f89f7a88

acceptorwrapper/rate_limiting_conn.go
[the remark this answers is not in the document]
felipejfc: I think that the only error possible is a broken pipe error, so it actually makes sense

acceptorwrapper/rate_limiting_conn.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/topfreegames/pitaya"
+	"github.com/topfreegames/pitaya/constants"
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/metrics"
+)
+
+// RateLimitingConn wraps net.Conn by applying rate limiting
+// and return empty if exceeded
+type RateLimitingConn struct {
```
felipejfc: nit: rename to RateLimiter

acceptorwrapper/rate_limiting_conn.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return n, err
+	}
+
+	now := time.Now()
+	err = r.take(now)
+	if err != nil {
+		logger.Log.Errorf("Data=%s, Error=%s", b, err)
+		metrics.ReportExceededRateLimiting(pitaya.GetMetricsReporters())
+		return 0, nil
```
felipejfc: I'm not sure you want to return here, the default behaviour of net.Conn is that it only returns when it has data to return, returning here you're changing this behaviour for the ones who are consuming a wrapped connection, isn't it better to have this inside a while and have it continue on errors?

Henrod: Ok, that makes sense

andrehp: But on the other hand, from the client's point of view the messages are simply getting lost, the client isn't notified about the rate limiting happening, this might lead to hard to detect bugs.

felipejfc: I'm not sure if the client should be notified, in what situations would a client hit the limiting?
1) he's cheating/trying to attack us
2) we inserted a but in clients code
in both case I think that it's irrelevant for us to answer anything for them, what do you guys think?

felipejfc: @codingllama: any thoughts?

Henrod: I don't think it would be hard to detect because it's reported using metricsReporters and logged as an error

acceptorwrapper/rate_limiter_test.go
```diff
@@ -0,0 +1,153 @@
...
+		before func()
+		err    error
+	}{
+		"test_can_take_on_first_time": {
+			before: func() {},
+			err:    nil,
+		},
+		"test_can_take_missing_one_to_limit": {
+			before: func() {
+				r.take(now)
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

acceptorwrapper/rate_limiter_test.go
```diff
@@ -0,0 +1,153 @@
...
+				r.take(now)
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

acceptorwrapper/rate_limiter_test.go
```diff
@@ -0,0 +1,153 @@
...
+			},
+			err: nil,
+		},
+		"test_can_take_when_oldest_request_expired": {
+			before: func() {
+				r.take(now.Add(-2 * interval))
```
golangcibot: Error return value of `r.take` is not checked (from `errcheck`)

acceptorwrapper/rate_limiting_conn.go
codingllama: I'm not sure I have enough context here. For example:

* Is dropping this data OK could it cause bugs?
* Does rate-limiting at the connection level implies a 1:1 relationship between a Read and RPC, or could this cause corrupted reads?
* Could we rate limit at a higher abstraction level? (eg, per RPC)

acceptorwrapper/rate_limiting_conn.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return n, err
+}
+
+// take saves the now as time taken or returns an error if
+// in the limit of rate limiting
+func (r *RateLimitingConn) take(now time.Time) error {
```
codingllama: nit: `take` is not a very descriptive name. I'd suggest something that alludes to rate limiting, maybe `shouldRateLimit() bool`?

acceptorwrapper/rate_limiting_conn.go
Henrod: - The effect of dropping would be a timeout error on the client side. So it depends on how the game deals with it. Maybe expliciting this behaviour on docs? So they can be prepared for this case. 
- Currently, Read can have more than one RPC but never less (we don't fragment packets now). Even though we can have more than one, we think it's a good approximation to rate limit based on Read
- Yes we could. But in my discussion with @felipejfc we thought it was better to not mix the logic afterwards and rate limit as soon as possible. Meaning that we don't need to decode the packets to rate limit it.

acceptorwrapper/rate_limiting_conn.go
Henrod: I agree about the name. 

Don't you think it's good to return the error here instead of only a bool?

codingllama: Either works, no strong opinion here. I took it out because it seems it only exists to be logged, in which case you could just write the message directly on `logger.Log.Errorf`.

Henrod: Fair enough. And `shouldRateLimit` makes more sense if returns a bool. I will change it.

Henrod:
@codingllama @felipejfc @cscatolini @andrehp any more thoughts?

cscatolini:
LGTM, but I think the GolangCI is still complaining about a few issues

I liked the approach we ended up with :)

Henrod:
> LGTM, but I think the GolangCI is still complaining about a few issues
> 
> I liked the approach we ended up with :)

The error on `metrics/report.go` we already don't check =(

acceptorwrapper/base.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) TFG Co. All Rights Reserved.
```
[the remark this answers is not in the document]
Henrod: Uhm, we don't have the year in almost any file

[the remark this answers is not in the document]
andrehp: =/
I think every file should contain every year it had a modification, but I don't know if modifying the whole project is worth it.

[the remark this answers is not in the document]
codingllama: I used to do year of conception. Copyright can be tricky, it may be worth looking for proper advice.

If the entire project is like this, then I'd suggest leaving it as-is here and maybe adding the year in a separate PR.

codingllama:
> @codingllama @felipejfc @cscatolini @andrehp any more thoughts?

No blockers from me, if the other reviewers are happy go for it. :)

PR #104, opened 2019-06-18
Properly closing grpc connections on deregister

andrehp:
The gRPC RPC client is currently leaking connections because they are not being properly closed.

References:
https://github.com/grpc/grpc-go/issues/2086
https://godoc.org/google.golang.org/grpc#ClientConn.NewStream


cluster/grpc_rpc_client.go
```diff
@@ -75,6 +75,11 @@ func NewGRPCClient(
 	return gs, nil
 }
 
+type grpcServer struct {
```
felipejfc: nit: doesn't make sense to call this struct grpcClient? and also the "srv" attribute is actually a client, sounds kinda strange...

cluster/grpc_rpc_client.go
```diff
@@ -148,14 +162,14 @@ func (gs *GRPCClient) BroadcastSessionBind(uid string) error {
 	}
 	fid, _ := gs.bindingStorage.GetUserFrontendID(uid, gs.server.Type)
 	if fid != "" {
-		if c, ok := gs.clientMap.Load(fid); ok {
+		if s, ok := gs.clientMap.Load(fid); ok {
```
felipejfc: c? :p

cluster/grpc_rpc_client.go
andrehp: I thought about it, my line of thought was that this struct contained information about a grpc server. But thinking now I tend to agree with you, I will change grpcServer to grpcClient and srv to cli

cluster/grpc_rpc_client.go
andrehp: Gonna change that as well

cluster/grpc_rpc_client.go
Henrod: We already have `GRPCClient` in this file. I would recommend the name `grpcConn`

PR #109, opened 2019-09-05
Lazy gRPC connections

andrehp:


cluster/grpc_rpc_client.go
```diff
@@ -106,7 +108,7 @@ func (gs *GRPCClient) Call(
 
 	parent, err := tracing.ExtractSpan(ctx)
 	if err != nil {
-		logger.Log.Warnf("failed to retrieve parent span: %s", err.Error())
+		logger.Log.Warnf("[grpc client] failed to retrieve parent span: %s", err.Error())
```
cscatolini: Shouldn't "grpcClient" be a field in the structured logging instead of a text?

andrehp: Yeah, ideally it should be, but pitaya's Logger interface doesn't expose the option to add fields :/

cscatolini: Oh noes,`[grpc client]` looks great then. Maybe something to improve in the future?

andrehp: Agreed

PR #114, opened 2019-10-23
[WIP] retry sending heartbeat

Henrod:
I want to discuss this fix. I don't know if make sense to retry. Before we were assuming that if failed to write on the connection, then the connection doesn't exist anymore. Is there a fact that this is always de case?

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -308,7 +308,8 @@ func (a *Agent) heartbeat() {
 				return
 			}
 			if _, err := a.conn.Write(hbd); err != nil {
-				return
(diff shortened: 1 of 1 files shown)


felipejfc:
Hi @Henrod I'm pretty sure that this will only return an error if the pipe is broken. So we shouldn't change the behavior.
Also, thinking about the problem we are facing now. It's not that we are receiving an error on sending the heartbeat, it's that the client is ignoring it.

Henrod:
Yes, I thought that this could be happening. I will close this PR then.

andrehp:
Actually, the heartbeat is not reaching the client

PR #115, opened 2019-10-25
Logging errors writing heartbeat

andrehp:


cscatolini:
@felipejfc what do you think about this PR? Should we merge it?

felipejfc:
This is an expected error. It should happen when the client is disconnected so I think we shouldn't. At least not as an error, maybe a debug log would be useful.

henriqueoelze:
Closing the pull request because of inactivity. If you judge that it is necessary, please open a new one.

PR #119, opened 2019-11-08
Add server type blacklist and concurrent server get on etcd to improv…

leohahn:
…e startup time.

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -453,7 +453,6 @@ func (p *parallelGetter) start() {
 				sv, err := getServerFromEtcd(p.cli, work.serverType, work.serverID)
 				if err != nil {
 					logger.Log.Errorf("error getting server from etcd: %s, error: %s", work.serverID, err.Error())
-					println("calling done")
 					p.wg.Done()
 					continue
 				}
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -281,9 +281,9 @@ func getKey(serverID, serverType string) string {
 	return fmt.Sprintf("servers/%s/%s", serverType, serverID)
 }
 
-func (sd *etcdServiceDiscovery) getServerFromEtcd(serverType, serverID string) (*Server, error) {
+func getServerFromEtcd(cli *clientv3.Client, serverType, serverID string) (*Server, error) {
 	svKey := getKey(serverID, serverType)
-	svEInfo, err := sd.cli.Get(context.TODO(), svKey)
+	svEInfo, err := cli.Get(context.TODO(), svKey)
 	if err != nil {
 		return nil, fmt.Errorf("error getting server: %s from etcd, error: %s", svKey, err.Error())
 	}
@@ -425,7 +425,7 @@ func (sd *etcdServiceDiscovery) processPendingServer(
 
 	svType, svID := sv.Type, sv.ID
 
-	sv, err := sd.getServerFromEtcd(svType, svID)
+	sv, err := getServerFromEtcd(sd.cli, svType, svID)
 	if err != nil {
 		logger.Log.Errorf("error getting server from etcd: %s, error: %s", svID, err.Error())
 		return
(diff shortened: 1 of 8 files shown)


cluster/etcd_service_discovery_test.go
```diff
@@ -367,3 +404,47 @@ func TestEtcdWatchChangesDeleteServers(t *testing.T) {
...
+				_, err := cli.Put(
+					context.TODO(),
+					getKey(serverToAdd.ID, serverToAdd.Type),
+					serverToAdd.AsJSONString(),
+				)
+				assert.NoError(t, err)
+			}
+
+			// We sleep to guarantee that the servers were added to etcd
+			time.Sleep(time.Millisecond * 50)
```
Henrod: This may make the test flaky, don't you think?

leohahn: The server is in memory, so this should be a very large margin. Do you think something else is required?

leohahn: Oh, the cli.Put method is synchronous, so the sleep is probably unnecessary.

cluster/etcd_service_discovery.go
```diff
@@ -108,6 +110,16 @@ func (sd *etcdServiceDiscovery) configure() {
 	sd.grantLeaseMaxRetries = sd.config.GetInt("pitaya.cluster.sd.etcd.grantlease.maxretries")
 	sd.grantLeaseInterval = sd.config.GetDuration("pitaya.cluster.sd.etcd.grantlease.retryinterval")
 	sd.shutdownDelay = sd.config.GetDuration("pitaya.cluster.sd.etcd.shutdown.delay")
+	sd.serverTypesBlacklist = sd.config.GetStringSlice("pitaya.cluster.sd.etcd.servertypeblacklist")
+	sd.syncServersParallelism = sd.config.GetInt("pitaya.cluster.sd.etcd.syncserversparallelism")
+
+	if sd.syncServersParallelism == 0 {
```
felipejfc: You should add all the configs here https://github.com/topfreegames/pitaya/blob/master/config/config.go
Also, by doing that you will not need this if, because you'll already specify the default value there.

cluster/etcd_service_discovery.go
```diff
@@ -108,6 +110,16 @@ func (sd *etcdServiceDiscovery) configure() {
...
+		sd.syncServersParallelism = 10
+	}
+
+	if len(sd.serverTypesBlacklist) > 0 {
+		logger.Log.Info("using server types blacklist: %s", sd.serverTypesBlacklist)
```
felipejfc: I'd prefer this to be a warning

cluster/etcd_service_discovery.go
```diff
@@ -405,6 +417,76 @@ func (sd *etcdServiceDiscovery) printServers() {
...
+
+				sv, err := getServerFromEtcd(p.cli, work.serverType, work.serverID)
+				if err != nil {
+					logger.Log.Errorf("error getting server from etcd: %s, error: %s", work.serverID, err.Error())
+					println("calling done")
+					p.wg.Done()
+					return
+				}
+
+				// We add into the resulting servers array
```
felipejfc: this comment is unnecessary

cluster/etcd_service_discovery.go
```diff
@@ -405,6 +417,76 @@ func (sd *etcdServiceDiscovery) printServers() {
...
+		go func() {
+			for work := range p.workChan {
+				logger.Log.Debugf("loading info from missing server: %s/%s", work.serverType, work.serverID)
+
+				sv, err := getServerFromEtcd(p.cli, work.serverType, work.serverID)
+				if err != nil {
+					logger.Log.Errorf("error getting server from etcd: %s, error: %s", work.serverID, err.Error())
+					println("calling done")
+					p.wg.Done()
+					return
```
felipejfc: do you really want to kill the goroutine in this case or just continue?

cluster/etcd_service_discovery.go
```diff
@@ -502,6 +598,19 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
 			select {
 			case wResp := <-chn:
 				for _, ev := range wResp.Events {
+					// we parse the string and check whether it is valid or not.
```
felipejfc: unnecessary comment

cluster/etcd_service_discovery.go
```diff
@@ -502,6 +598,19 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
...
+					svType, svID, err := parseEtcdKey(string(ev.Kv.Key))
+					if err != nil {
+						logger.Log.Warnf("failed to parse key from etcd: %s", ev.Kv.Key)
+						continue
+					}
+
+					// we then check if the server type is blacklisted or not
```
felipejfc: unnecessary comment

cluster/nats_rpc_client.go
```diff
@@ -79,6 +81,9 @@ func (ns *NatsRPCClient) configure() error {
 	if ns.connString == "" {
 		return constants.ErrNoNatsConnectionString
 	}
+	if timeout := ns.config.GetDuration("pitaya.cluster.rpc.client.nats.connectiontimeout"); timeout != 0 {
```
felipejfc: add this to config.go as well

cluster/nats_rpc_server.go
```diff
@@ -88,6 +91,9 @@ func (ns *NatsRPCServer) configure() error {
 	if ns.connString == "" {
 		return constants.ErrNoNatsConnectionString
 	}
+	if timeout := ns.config.GetDuration("pitaya.cluster.rpc.server.nats.connectiontimeout"); timeout != 0 {
+		ns.connectionTimeout = timeout
```
felipejfc: same as above

felipejfc:
Next time please split the PR, e.g.

PR1 - Server blacklist
PR2 - Concurrent server gets

cluster/etcd_service_discovery.go
leohahn: good point, it should be a continue

cluster/etcd_service_discovery.go
```diff
@@ -530,3 +631,12 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
 		}
 	}(w)
 }
+
+func (sd *etcdServiceDiscovery) isServerTypeBlacklisted(svType string) bool {
+	for _, blacklistedSv := range sd.serverTypesBlacklist {
```
felipejfc: Isn't this going to crash if sd.serverTypesBlacklist is nil? (I saw that it's the default value in the config)

leohahn: No, a nil slice is considered an empty slice in go.
https://play.golang.org/p/GgbP0zUX8kP

cluster/etcd_service_discovery_test.go
felipejfc: there are better ways of waiting in a test, something like that:

https://github.com/topfreegames/pitaya/blob/master/cluster/nats_rpc_client_test.go#L158

let me know what you think

felipejfc: other example https://github.com/topfreegames/pitaya/blob/931f2880440b52b045c553bdb0b20502fdba09bc/session/session_test.go#L612

leohahn: there is no simple way of doing that because for some reason `sync.Map` does not have a Len method.  But I'll see a way of doing it.

cluster/etcd_service_discovery_test.go
```diff
@@ -434,18 +434,34 @@ func TestEtcdWatchChangesWithBlacklist(t *testing.T) {
...
+				if err == nil {
+					num++
+				}
+				for _, serverToAdd := range table.serversToAdd {
+					_, err := e.GetServer(serverToAdd.ID)
+					if err == nil {
+						num++
+					}
 				}
+				println("Returning val ", num)
```
felipejfc: delete this line

cluster/etcd_service_discovery.go
felipejfc: Why do we still have this if? Aren't you specifying a default for that already?

PR #124, opened 2019-12-14
feat: read bigger messages

felipejfc:
We have a limit of 16MB for client messages, however, we had a hard-limit for receiving 2kb data chunks from the clients. This means that any message > 2kb would cause the client to be disconnected. In this commit I make it possible for the server to read chunks that are up-to MaxPacketSize(16mb) in length.

### DIFF
--- a/service/handler.go
+++ b/service/handler.go
@@ -180,7 +180,7 @@ func (h *HandlerService) Handle(conn net.Conn) {
 	}()
 
 	// read loop
-	data := make([]byte, 4096)
+	data := make([]byte, constants.IOBufferBytesSize)
 	buf := bytes.NewBuffer(nil)
 	for {
 		totalLen := 0
--- a/service/handler.go
+++ b/service/handler.go
@@ -21,6 +21,7 @@
 package service
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


recklessMo:
@felipejfc   

I believe that the `totalLen` has nothing to do with the `MaxPacketSize`.  If the client does something as below, the server may receive a series of logical packets in just one read operation.

```
	data := "0123456789abcd"
	i := 0
	for i < 10{
		n, err := conn.Write([]byte(data))
		if err != nil {
			fmt.Println("write errors", err)
			break
		}
		fmt.Println("client send: ", n)
		i++
	}
```

felipejfc:
@recklessMo you're right, but the goal is to put some kind of limit in the amount of data the client sends to the server to avoid flooding attacks.
I just used MaxPacketSize because I think 16mb is enough if someone is sending more than that at once from a client than probably they're doing something wrong.

service/handler.go
```diff
@@ -179,25 +180,36 @@ func (h *HandlerService) Handle(conn net.Conn) {
...
 	for {
-		n, err := conn.Read(buf)
-		if err != nil {
-			logger.Log.Debugf("Read message error: '%s', session will be closed immediately", err.Error())
-			return
+		totalLen := 0
+		n := len(data)
+		var err error
+		for n == len(data) {
+			n, err = conn.Read(data)
```
leohahn: What happens if Read returns one and a half packets? I think currently it would throw the second packet away. Shouldn't we shift the remaining bytes to the beginning of the buffer and append with the next read call? (I don't know if that happens in practice, in theory I think it should)

leohahn: I guess this issue somewhat relates to what @recklessMo said

PR #125, opened 2019-12-25
Refactor player connection

### LINKED ISSUE #123: question about the packet decoder
https://github.com/topfreegames/pitaya/blob/287b60e35ecae869b29d87df0b782969b4bf0c09/service/handler.go#L193

If a client sends  a packet of 1500 bytes,  but the underlying connection only read 1000 bytes for the first read attempt and for the second attempt it read the remaining 500 bytes, which I think is possible.

The referring code will cause decode error because it doesnot have enough bytes and it will discard the first 1000 bytes.  how to resolve this case ？

felipejfc:
This PR decouples the logic of reading messages from the service/handler and delegates the responsibility for the acceptor itself by implementing a GetNextMessage method. TCP Acceptor now parses whole messages one by one from the stream, this way we ensure not to read partial messages anymore.

This fixes #123


agent/agent.go
```diff
@@ -103,8 +103,14 @@ func NewAgent(
 	metricsReporters []metrics.Reporter,
 ) *Agent {
 	// initialize heartbeat and handshake data on first user connection
+	serializerName := ""
+
+	if serializer != nil { //should never be true, only during testing
```
cscatolini: I don't like this approach. IMO we'd better ensure serializer is not `nil` on the tests causing this issue instead of inserting this clause.

service/handler.go
```diff
@@ -220,7 +205,7 @@ func (h *HandlerService) Handle(conn net.Conn) {
 				return
 			}
 		}
-		buf.Reset()
+		//buf.Reset()
```
cscatolini: Remove this line?

agent/agent.go
felipejfc: I was lazy when coding this, I also prefer this other approach!

PR #127, opened 2020-01-06
Refactor Agent

### LINKED ISSUE #112: concurrent write to websocket connection
concurrent write to websocket connection
```
	if c.isWriting {
		panic("concurrent write to websocket connection")
	}
	c.isWriting = true
```

when use websocket, we find error code. agent use two goroutine for write websocket conn. but websocket ntoes : Connections support one concurrent reader and one concurrent writer
```
func (a *Agent) Handle() {
	defer func() {
		a.Close()
		logger.Log.Debugf("Session handle goroutine exit, SessionID=%d, UID=%d", a.Session.ID(), a.Session.UID())
	}()

	go a.write()
	go a.heartbeat()
	select {
	case <-a.chDie: // agent closed signal
		return
	}
}
```

https://godoc.org/github.com/gorilla/websocket#hdr-Concurrency
Concurrency ¶
Connections support one concurrent reader and one concurrent writer.
Applications are responsible for ensuring that no more than one goroutine calls the write methods (NextWriter, SetWriteDeadline, WriteMessage, WriteJSON, EnableWriteCompression, SetCompressionLevel) concurrently and that no more than one goroutine calls the read methods (NextReader, SetReadDeadline, ReadMessage, ReadJSON, SetPongHandler, SetPingHandler) concurrently.
The Close and WriteControl methods can be called concurrently with all other methods.

so we must fix agent.go， use one goroutine to send websocket msg like nano

felipejfc:
With this pr we send writes from only 1 goroutine (fixes #112), also, make the logic more readable.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -61,12 +61,12 @@ const handlerType = "handler"
 type (
 	// Agent corresponds to a user and is used for storing raw Conn information
 	Agent struct {
-		Session            *session.Session    // session
-		appDieChan         chan bool           // app die channel
-		chDie              chan struct{}       // wait for close
-		chSend             chan pendingMessage // push message queue
-		chStopHeartbeat    chan struct{}       // stop heartbeats
-		chStopWrite        chan struct{}       // stop writing messages
+		Session            *session.Session  // session
+		appDieChan         chan bool         // app die channel
+		chDie              chan struct{}     // wait for close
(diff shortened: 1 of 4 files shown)


felipejfc:
@cscatolini I'd like your review since you were the one who implemented jaeger and I made some changes to reporting hoping not to change the logic :)

PR #133, opened 2020-07-13
Refactor static API to improve Pitaya testability

felippeduran:
In summary, this PR refactors the entire `pitaya` package API to make it non static.

The goal of this change is to make pitaya more testable internally, and also by applications using it. Instead of implementing wrappers, applications can directly use the mocked implementations from pitaya for tests.

In order to do that, several static APIs from many packages (`component`, `modules`, `pipeline`, `groups`, `sessions` and `remote`) were moved to non static also, removing static variables and moving them to instances referenced by their users. Container classes were generated from them, like `SessionPool` and `HandlerHooks`. Finally, completely new classes were created, like the `agentFactoryImpl` and `Builder` to create interfaces instead of concrete implementations.

Interfaces created:
* `PitayaBuilder`
* `Pitaya`
* `Agent`, `AgentFactory`
* `Session`, `SessionPool`

Renames:
* Old `Agent` was renamed to `agentImpl`
* Old `Session` was renamed to `sessionImpl`

All tests and decumentation were changed to match the new API. Commands were added to the Makefile in order to generate mocks for new interfaces.

Although this PR doesn't make structural behaviour changes, it breaks code compatibility with previous versions, so it's recommend for it to be released as a major version.


Makefile
```diff
@@ -138,6 +138,18 @@ test-coverage-func coverage-func: test-coverage merge-profiles
 	@echo "\033[1;34m=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-\033[0m"
 	@go tool cover -func=coverage-all.out | egrep -v "100.0[%]"
 
+agent-mock:
```
leohahn, line 141: We generally only create one target for mocking.

felippeduran: I usually do the same, but I ended up following the format already in place. There's a specific target for the `Serializer` and the `Acceptor`

leohahn: Oh ok then.

agent/agent.go
```diff
@@ -94,10 +94,76 @@ type (
...
+		SetStatus(state int32)
+		Handle()
+		IPVersion() string
+		SendHandshakeResponse() error
+		SendRequest(ctx context.Context, serverID, route string, v interface{}) (*protos.Response, error)
+		AnswerWithError(ctx context.Context, mid uint, err error)
+	}
+
+	// AgentFactory factory for creating Agent instances
+	AgentFactory interface {
```
leohahn, line 118: What is the rationale for creating a factory here instead of the instance directly?

agent/agent_test.go
```diff
@@ -92,9 +92,10 @@ func TestNewAgent(t *testing.T) {
 	messageEncoder := message.NewMessagesEncoder(false)
 
 	mockMetricsReporters := []metrics.Reporter{mockMetricsReporter}
+	sessionPool := session.NewSessionPool()
 
 	mockMetricsReporter.EXPECT().ReportGauge(metrics.ConnectedClients, gomock.Any(), gomock.Any())
-	ag := NewAgent(mockConn, mockDecoder, mockEncoder, mockSerializer, hbTime, 10, dieChan, messageEncoder, mockMetricsReporters)
+	ag := newAgent(mockConn, mockDecoder, mockEncoder, mockSerializer, hbTime, 10, dieChan, messageEncoder, mockMetricsReporters, sessionPool).(*agentImpl)
```
leohahn, line 98: maybe `newAgentImpl` instead of `newAgent` would be better as a function name.

leohahn: Actually, newAgent could return the concrete type instead of the interface, that way you wouldn't need to cast the type here.

app.go
```diff
@@ -92,241 +146,132 @@ type App struct {
...
-
-func startDefaultRPCClient() {
-	// initialize default rpc client
-	rpcClient, err := cluster.NewNatsRPCClient(app.config, app.server, app.metricsReporters, app.dieChan)
-	if err != nil {
-		logger.Log.Fatalf("error starting cluster rpc client component: %s", err.Error())
-	}
-	SetRPCClient(rpcClient)
+// SetLogger logger setter
+func SetLogger(l logger.Logger) {
```
leohahn, line 262: Why is this one still static?

app.go
```diff
@@ -92,241 +146,132 @@ type App struct {
...
+	serializer serialize.Serializer,
+	acceptors []acceptor.Acceptor,
+	dieChan chan bool,
+	router *router.Router,
+	server *cluster.Server,
+	rpcClient cluster.RPCClient,
+	rpcServer cluster.RPCServer,
+	worker *worker.Worker,
+	serviceDiscovery cluster.ServiceDiscovery,
+	remoteService *service.RemoteService,
```
leohahn, line 171: Why are you passing this strcut on the constructor instead of creating it internally how it was before?

app.go
```diff
@@ -521,13 +414,13 @@ func Error(err error, code string, metadata ...map[string]string) *errors.Error
 }
 
 // GetSessionFromCtx retrieves a session from a given context
-func GetSessionFromCtx(ctx context.Context) *session.Session {
+func (app *App) GetSessionFromCtx(ctx context.Context) session.Session {
```
leohahn, line 417: Does this need to be a method? It does not use the receiver.

builder.go
```diff
@@ -0,0 +1,248 @@
...
+	ServerMode       ServerMode
+	ServiceDiscovery cluster.ServiceDiscovery
+	Groups           groups.GroupService
+	SessionPool      session.SessionPool
+	Worker           *worker.Worker
+	HandlerHooks     *pipeline.HandlerHooks
+}
+
+// PitayaBuilder Builder interface
+type PitayaBuilder interface {
```
leohahn, line 48: Does the builder needs to be an interface as well?

felippeduran: Good point, it doesn't. I probably forgot to remove it.
Also, btw, I just created this builder to keep functionality as it was before, easily creating pitaya with lots of default stuff, but also giving flexibility to customize implementations.

In general, I don't like this "default" approach, and in my use case I ended up not using it. So we might want to consider removing the builder altogether and using the explicit NewApp constructor with all required dependencies.

The downside is having to explicitly instantiate all dependencies.

app.go
felippeduran: That's a good question. I didn't make the change because it wasn't critical for my end objective, which was making pitaya more testable.
Also, I didn't want to make another shotgun surgery. Maybe we can do this in a later PR.

examples/demo/worker/services/worker.go
```diff
@@ -15,18 +15,15 @@ type Worker struct {
...
-func (w *Worker) Configure() error {
-	err := pitaya.StartWorker(pitaya.GetConfig())
-	if err != nil {
-		return err
-	}
-
-	pitaya.RegisterRPCJob(&RPCJob{})
-	return nil
+func (w *Worker) Configure(app pitaya.Pitaya) {
+	app.StartWorker()
```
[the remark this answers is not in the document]
felippeduran: Because the error could be generated by config parsing. As it's not being configured here, there's no more errors that could occur in this place. Now, it's being handled in the worker constructor.
This is the old snippet:
```go
	hostname, err := os.Hostname()
	if err != nil {
		return nil, err
	}
```

helpers/helpers.go
```diff
@@ -153,6 +153,7 @@ func StartServer(
 	}
 
 	err = cmd.Start()
+	fmt.Println(err)
```
leohahn, line 156: Probably shouldn't be here

felippeduran: Oh, right! Will remove it!

felippeduran: Done

app.go
felippeduran: The idea is to avoid deep construction hierarchies. The deeper they are, the harder it is to refactor stuff. Also, it's easier to mock dependencies.

The alternative would be to access private fields or create public getters just for this purpose. And this last one would make possible to change it after it's creation, which could cause unpredictable behaviours and more edge cases to test.

service/base_service.go
```diff
@@ -0,0 +1,11 @@
+package service
+
+import "github.com/topfreegames/pitaya/pipeline"
+
+type baseService struct {
```
leohahn, line 5: What's the purpose of this struct?

app.go
felippeduran: That's a good question. It indeed doesn't need to be in practice. On the other hand, this is required to properly mock it. You can check in the refactored tests that I'm doing that.

felippeduran:
@leohahn, I've seen that it's a common practice to create a /v2 path with an independent module for major changes. However it duplicates the entire codebase. `go-pg` seems to ignore this approach completely, for example. I don't know how you guys want to proceed about this here.

session/session_test.go
```diff
@@ -445,51 +464,53 @@ func TestSessionBindRunsOnSessionBind(t *testing.T) {
...
-	ss := New(nil, true)
+	sessionPool := NewSessionPool().(*sessionPoolImpl)
+	ss := sessionPool.NewSession(nil, true)
 	assert.NotNil(t, ss)
 
 	uid := uuid.New().String()
 	err := ss.Bind(nil, uid)
 	assert.NoError(t, err)
-	assert.Equal(t, uid, ss.uid)
+	assert.Equal(t, uid, ss.UID())
```
leohahn, line 511: These interfaces are a little bit weird. They expose all of the things that the old struct did, so the only reason for them to exist is for testability purposes, which is not ideal.

service/handler_pool.go
```diff
@@ -0,0 +1,125 @@
...
+	e "github.com/topfreegames/pitaya/errors"
+	"github.com/topfreegames/pitaya/logger"
+	"github.com/topfreegames/pitaya/pipeline"
+	"github.com/topfreegames/pitaya/route"
+	"github.com/topfreegames/pitaya/serialize"
+	"github.com/topfreegames/pitaya/session"
+	"github.com/topfreegames/pitaya/util"
+)
+
+// HandlerPool ...
```
leohahn, line 20: Can you comment the purpose of this struct?

service/handler.go
```diff
@@ -69,6 +68,8 @@ type (
 		services         map[string]*component.Service // all registered service
 		metricsReporters []metrics.Reporter
 		agentFactory     agent.AgentFactory
+		handlerPool      *HandlerPool
+		handlers         map[string]*component.Handler // all handler method
```
leohahn, line 72: Why do you have another handlers map here since you already have a pool?

service/handler_pool_test.go
```diff
@@ -0,0 +1,194 @@
...
+		errSerReturn error
+		errSerialize error
+		outSerialize interface{}
+		handlerType  message.Type
+		msgType      interface{}
+		remote       bool
+		out          []byte
+		err          error
+	}{
+		{"invalid_route", route.NewRoute("", "no", "no"), nil, nil, nil, message.Request, nil, false, nil, e.NewError(errors.New("pitaya/handler: no.no not found"), e.ErrNotFoundCode)},
```
leohahn, line 100: Can you break lines here?

app.go
leohahn: Cool.

builder.go
leohahn: I think having a builder is still desirable because it allows you to quickly create a server for testing/prototyping. Since it is an open-source project, we should have an easy way to create a server.

leohahn:
Can you move this merge to another branch, like v2? This follows the talk we had.

felippeduran:
> Can you move this merge to another branch, like v2? This follows the talk we had.

Done, @leohahn.

PR #134, opened 2020-07-14
Refactor/hardcoded configs

felippeduran:
This PR depends on #133. It's recommended to review that one first so the diff in this one gets reduced.

This PR creates explicit config structs for creating object instances. It removes spread, hardcoded configs from all over the code and move them to specific config constructor methods. In the future, they can be completely removed, leaving configuration as a responsibility of the app using pitaya. It also adds default constructors to not lose the functionality of default configs for running pitaya out-of-the-box.


felippeduran:
@leohahn, this PR currently have an approach of leaving config constructors within their corresponding package. However, this leads to spread out default configuration values. What do you think about this? I've discussed this approach with @felipejfc, but I don't quite remember deciding for any specific direction.

leohahn:
> @leohahn, this PR currently have an approach of leaving config constructors within their corresponding package. However, this leads to spread out default configuration values. What do you think about this? I've discussed this approach with @felipejfc, but I don't quite remember deciding for any specific direction.

Won't it be better if we have a package with all of the default values? I agree that it is bad to have default values all over the place.

felippeduran:
> > @leohahn, this PR currently have an approach of leaving config constructors within their corresponding package. However, this leads to spread out default configuration values. What do you think about this? I've discussed this approach with @felipejfc, but I don't quite remember deciding for any specific direction.
> 
> Won't it be better if we have a package with all of the default values? I agree that it is bad to have default values all over the place.

Done!

cluster/config_info_retriever.go
```diff
@@ -9,9 +9,9 @@ type ConfigInfoRetriever struct {
 }
 
 // NewConfigInfoRetriever returns a *ConfigInfoRetriever
-func NewConfigInfoRetriever(c *config.Config) *ConfigInfoRetriever {
+func NewConfigInfoRetriever(config config.ConfigInfoRetrieverConfig) *ConfigInfoRetriever {
```
leohahn, line 12: since the package is already called `config`, can't you remove the `Config` prefix from the struct name?

leohahn:
So, overall LGTM, but I'm still not 100% sold on the config package. The reason is that the packages are not self contained anymore. For example, the etcd struct now depends on the config package that contains information for all of the other things on the project.

I think that ideally the config package would only aggregate the types form other packages into a larger `Config` struct. The default values would then be defined alone by themselves on the config package or in a constants package. I'm doing something like this here: https://github.com/tfgco/pitaya-rs/blob/master/src/pitaya/src/settings.rs

But I'd like to know your opinion on this as well.

felippeduran:
> So, overall LGTM, but I'm still not 100% sold on the config package. The reason is that the packages are not self contained anymore. For example, the etcd struct now depends on the config package that contains information for all of the other things on the project.
> 
> I think that ideally the config package would only aggregate the types form other packages into a larger `Config` struct. The default values would then be defined alone by themselves on the config package or in a constants package. I'm doing something like this here: https://github.com/tfgco/pitaya-rs/blob/master/src/pitaya/src/settings.rs
> 
> But I'd like to know your opinion on this as well.

I originally went for this approach, having the config structs defined in the original package and only the factory methods implemented within the config package. The problem with this is that it creates a dependency cycle when implementing tests, which Golang doesn't allow.

The alternative I see is to create a sub package for defining config structs. Then, the config package would reference this one, and tests would reference the config struct sub package and the config one, avoiding the cycle. I ended up not going with this approach to avoid spending more time in this specific PR, but I can make the change if you prefer. What do you think?

About specifically having a larger Config struct to aggregate everything, I don't think it's that useful in this case as most of the time configs for different packages are used separately. Also, some of them are mutually exclusive, like Nats and Grpc, thus not making sense to have them live simultaneously within this larger Config.

leohahn:
> > So, overall LGTM, but I'm still not 100% sold on the config package. The reason is that the packages are not self contained anymore. For example, the etcd struct now depends on the config package that contains information for all of the other things on the project.
> > I think that ideally the config package would only aggregate the types form other packages into a larger `Config` struct. The default values would then be defined alone by themselves on the config package or in a constants package. I'm doing something like this here: https://github.com/tfgco/pitaya-rs/blob/master/src/pitaya/src/settings.rs
> > But I'd like to know your opinion on this as well.
> 
> I originally went for this approach, having the config structs defined in the original package and only the factory methods implemented within the config package. The problem with this is that it creates a dependency cycle when implementing tests, which Golang doesn't allow.
> 
> The alternative I see is to create a sub packages for defining config structs. Then, the config package would reference them, and tests would reference both, avoiding the cycle. I ended up not going with this approach to avoid spending more time in this specific PR, but I can make the change if you prefer. What do you think?
> 
> About specifically having a larger Config struct to aggregate everything, I don't think it's that useful in this case as most of the time configs for different packages are used separately. Also, some of them are mutually exclusive, like Nats and Grpc, thus not making sense to have them live simultaneously within this larger Config.

Yeah, I guess the current solution is reasonable then. I think the refactor you mentioned is not really necessary, it looks like it adds more complexity which is not really worth it. Overall , LGTM then.

PR #135, opened 2020-07-16
Improve Logger interface and create wrapper for logrus.Logger

felippeduran:
This PR improves decoupling of specific logging implementations from pitaya to allow easy swapping of loggers. Ideally, this will allow removing logrus altogether from the repo and leaving to the user's choice a logging solution.

It adds some new methods to the `Logger` interface, moves it to a `logger/interfaces` package and add a `logrus` package with a wrapper implementation.


client/client.go
```diff
@@ -97,7 +98,7 @@ func New(logLevel logrus.Level, requestTimeout ...time.Duration) *Client {
 	l.Formatter = &logrus.TextFormatter{}
 	l.SetLevel(logLevel)
 
-	logger.Log = l
+	logger.Log = logruswrapper.NewWithLogger(l)
```
danielbdias, line 101: Thinking about the encapsulation of the logging interface, makes sense to create a Factory to build these loggers (along with formatter and loglevel config)? The main (and costly) issue to do that is how to encapsulate log configuration too.
 
Doing that we could centralize all calls to `logruswrapper.NewWithLogger` there avoiding logger users to know that the real interface is implemented by `logrus`. 

What do you think?

felippeduran: I think it makes sense, but only if we decide to go in the opposite direction of what I'm proposing here.

My idea is to completely remove the dependency from logrus or any specific logger implementation. In this scenario, we would remove `logruswrapper` from the code base. Therefore, I think the ideal change would be to receive the abstract `interfaces.Logger` as a constructor parameter. And having no logging implementations, then it doesn't make sense to have a factory.

felippeduran: Oh, and just a side note, `logruswrapper` already have a factory method, `logruswrapper.New()`. This alternative `NewWithLogger()` is intended to allow complete logrus configuration without having to create a factory class duplicating all configurability. Because in the end, the more we expose as configuration, the more we'll bind code to the actual logrus implementation, making the case for just having `NewWithLogger()` as we have now.

felippeduran:
Oh, just one note. This PR changes a tiny bit the library public API, so you might consider changing the base branch to v2.

client/client.go
danielbdias: Makes sense. Looking through pitaya logger I think that now with these changes we have a good isolation from `logrus` concrete implementations. 

My concern was because of two reasons:
1. Here, because of `logLevel` argument we are leaking `logrus` knowledge to outside users (I know this method is called only tests, so it's minor concern). 
2. All places that calls `NewWithLogger` has the same `init logrus / set json formatter / set log level` structure before calling it, and with exception of the logLevel logic they could be encapsulated in the factory.

felippeduran:
> Oh, just one note. This PR changes a tiny bit the library public API, so you might consider changing the base branch to v2.

@leohahn, I've moved the target branch to v2 and fixed all conflicts. It should be ready to merge now.

PR #136, opened 2020-07-28
Add network metrics to pitaya client

RonnanSouza:
# Network Metrics 

I implemented a struct called Metrics that holds Ping (round trip time), Jitter, and requests information, such as:

## Ping
* Current Ping that is the average of best two pings from the last three collected to avoid some spike
* LifetimePingPercentile to where is possible to get an overview from the lifetime ping measurements in this connection, similar to [this](https://github.com/ValveSoftware/GameNetworkingSockets/blob/f3f9c853d45d9f5aaa98f7eb53d895d514c57f19/src/common/percentile_generator.h).
## Jitter
* Current Jitter, that is the absolute difference from the last two collected pings
* LifetimeJitterPercentile similar to *LifetimePingPercentile*
## Requests
* Num of Requests sent
* Num of Requests acknowledged by the server
* Num of Requests not acknowledged by the server

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -79,6 +79,7 @@ type Client struct {
 	nextID              uint32
 	messageEncoder      message.Encoder
 	clientHandshakeData *session.HandshakeData
+	Metrics             *Metrics
 }
 
 // MsgChannel return the incoming message channel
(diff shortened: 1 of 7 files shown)


henriqueoelze:
Hello @RonnanSouza , thank you for the request. Can you give us some examples why these metrics are necessary on the client and an use case for them?

RonnanSouza:
Hey @henriqueoelze, it was from an old UFCG project that we were developing some time ago, but now it makes no sense anymore, please disregard this.

PR #137, opened 2020-08-07
Update util Pcal method to log stack trace as a raw string literal

felipeap92:
Changed how util Pcal method logs a stack trace when recovers from a panic error.


leohahn:
LGTM

leeohaddad:
LGTM

PR #139, opened 2020-09-04
Fix log print uid fmt string %d -> %s or %v.

Colstuwjx:
After run `make run-cluster-grpc-example-connector`, I see the log output below:

```
DEBU[9360] type: room, servers: map[fe10d5d3-7dde-4a06-b0d1-e1a51bcc7ea3:0xc0002ee240]  source=pitaya
ERRO[9435] Failed to process remote: session is already bound to an uid  source=pitaya
DEBU[9435] Type=Response, ID=4, UID=%!d(string=1), MID=2, Data=89bytes  source=pitaya
ERRO[9452] Failed to process remote: session is already bound to an uid  source=pitaya
DEBU[9452] Type=Response, ID=4, UID=%!d(string=1), MID=3, Data=89bytes  source=pitaya
```

It seems that we're using wrong format string for printing uid, so I proposed this PR for fixing this.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -215,10 +215,10 @@ func (a *Agent) Push(route string, v interface{}) error {
 
 	switch d := v.(type) {
 	case []byte:
-		logger.Log.Debugf("Type=Push, ID=%d, UID=%d, Route=%s, Data=%dbytes",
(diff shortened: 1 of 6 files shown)


Colstuwjx:
The travis CI error seems not related to this PR changes, so how could I fire retest?

felipejfc:
thanks for the fix!

PR #146, opened 2020-11-11
Refactor configs to allow unmarshalling instead of accessing them by string paths

felippeduran:
This PR is separated into two isolated commits:
* The first one refactors configs into nested structs to allow unmarshalling instead of accessing them by string paths
* The second one standardizes constructors to always return pointers instead of values (more aligned with the common use of the `new` keyword)
It's recommended to review them independently.

There's an [open issue](https://github.com/spf13/viper/issues/761) in [viper](https://github.com/spf13/viper) related to the use of env var overrides and unmarshalling, but one of the suggested workarounds is already covered by our code [here](https://github.com/topfreegames/pitaya/blob/v2/config/viper_config.go#L147).


builder.go
```diff
@@ -73,16 +73,16 @@ func NewBuilderWithConfigs(
...
-		customMetrics,
-		prometheusConfig,
-		statsdConfig,
-		etcdSDConfig,
-		natsRPCServerConfig,
-		natsRPCClientConfig,
-		workerConfig,
-		enqueueOpts,
-		groupServiceConfig,
+		*builderConfig,
```
rodopoulos, line 76: Something is off to me and I this might be a stupid statement: why don't you just use pointers along you struct instead of passing everything with pointers? Does Viper need you to declare it that way? Using pointers in the struct directly simplifies the code.

rodopoulos (changes requested):

config/viper_config.go
```diff
@@ -86,60 +84,62 @@ func (c *Config) fillDefaultValues() {
...
-		"pitaya.cluster.sd.etcd.syncservers.interval":           etcdSDConfig.SyncServersInterval,
-		"pitaya.cluster.sd.etcd.shutdown.delay":                 etcdSDConfig.ShutdownDelay,
+		"pitaya.cluster.rpc.server.nats.services":               natsRPCServerConfig.Services,
+		"pitaya.cluster.rpc.server.nats.buffer.messages":        natsRPCServerConfig.Buffer.Messages,
+		"pitaya.cluster.rpc.server.nats.buffer.push":            natsRPCServerConfig.Buffer.Push,
+		"pitaya.cluster.sd.etcd.dialtimeout":                    etcdSDConfig.DialTimeout,
+		"pitaya.cluster.sd.etcd.endpoints":                      etcdSDConfig.Endpoints,
+		"pitaya.cluster.sd.etcd.prefix":                         etcdSDConfig.Prefix,
+		"pitaya.cluster.sd.etcd.grantlease.maxretries":          etcdSDConfig.GrantLease.MaxRetries,
+		"pitaya.cluster.sd.etcd.grantlease.retryinterval":       etcdSDConfig.GrantLease.Interval,
```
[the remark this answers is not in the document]
felippeduran: Agreed!

config/config.go
```diff
@@ -274,7 +402,7 @@ func NewDefaultCustomMetricsSpec() models.CustomMetricsSpec {
 func NewCustomMetricsSpec(config *Config) models.CustomMetricsSpec {
 	var spec models.CustomMetricsSpec
 
-	err := config.UnmarshalKey("pitaya.metrics.custom", &spec)
+	err := config.UnmarshalKey("&pitaya.metrics.custom", &spec)
```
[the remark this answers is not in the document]
felippeduran: > Is this & in the key right?

No, terrible typo! Thanks for pointing that out.

> Also, consider maintaining code consistency, defining the err in the if block:

Sure! I ended up not noticing this as this line was already here before. Will change that.

builder.go
felippeduran: This is something that I was indeed in doubt about.

Before this MR, I was using values instead of pointers to make it explicit that configs are readonly for receivers (you always use a copy). And to make code simpler, constructors would return the value.

But then, it was annoying me that constructors returning values were breaking consistency across the `new` keyword common sense, and also within the library itself. Therefore I decided to stardardize to all of them return pointers, which would then require dereferencing then when passing these configs to functions that use them in that "readonly" format.

Just simply using pointers everywhere does indeed solves this conversion issue across the codebase, but it also goes against the "readonly" mindset that I think it's important in the first place.

What's your thought on this, @leohahn? Which way should we go?

leohahn: For me, it makes sense that config is a pointer. By having a pointer, you specify that there is only one instance of a config, which makes sense. For, the fact that Config is named that way is already implied that it is read-only, we don't need to pass values around.

I don't like that go does not have a const modified to specify that, since by passing values around, you can still modify the variable.

felippeduran: > By having a pointer, you specify that there is only one instance of a config

This is a very unsafe, implicit assumption right? There's absolutely no guarantee that this is actually the case 🤔. You wouldn't make the same statement in a C++ codebase, would you?

But as I'm the outsider here, then I can disagree and commit nevertheless to make the change 👍.
Should I replace everything by pointers then?

leohahn: The problem of comparing C++ and Go, is that C++ has `const`, which go does not. A config in C++ I would pass probably with `std::shared_ptr`, which is the equivalent of Golang's pointer. 

Also, if you receive things by value you still have some edge cases. For example, if there is any pointer inside the struct you copied, you still have the same issue, where the pointer inside the struct can be modified.

felippeduran: > The problem of comparing C++ and Go, is that C++ has const, which go does not. A config in C++ I would pass probably with std::shared_ptr, which is the equivalent of Golang's pointer.

I still don't agree, as there are many situations where you have `std::shared_ptr`s to different object instances being used in code. Figure like two player objects, two item instances, or even two matches running in a server, each one with different configurations.

> Also, if you receive things by value you still have some edge cases. For example, if there is any pointer inside the struct you copied, you still have the same issue, where the pointer inside the struct can be modified.

That's why I made all struct configs be referenced by value when nested into other configs 👍

Anyway, I let you know once I update the PR to always use pointers!

leohahn: >That's why I made all struct configs be referenced by value when nested into other configs 👍

fair enough!

gabrielcorado (approved):

luizmiranda7 (approved):

PR #148, opened 2020-12-04
Fix agent race condition

victor-carvalho:
When agent write goroutine closed the channel heartbeat and send functions could still write to the channel causing a panic. send already had a recover for this but heartbeat didn't have. This was causing multiple restarts on some applications using pitaya.

This commit does not close the channel anymore and to make so the writers don't block sending when an agent is closed a select is used reading from die channel that signals that an agent is closed. The write channel is never closed but it's ok as when a Agent is not used anymore it will be garbage collected as the agent is the only reference to it.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -203,7 +203,11 @@ func (a *Agent) send(pendingMsg pendingMessage) (err error) {
 		pWrite.err = util.GetErrorFromPayload(a.serializer, m.Data)
 	}
 
-	a.chSend <- pWrite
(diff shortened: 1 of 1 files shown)


rodopoulos (approved):
LGTM

leohahn:
LGTM, it would be awesome if we had a regression test for this bug!

victor-carvalho:
> LGTM, it would be awesome if we had a regression test for this bug!

I agree with you but I really don't know how to test this.

Issue #138, opened 2020-08-31
Connector logs "Error reading next available message: invalid header" on player disconnection


renatoaf:
**Version used:** v1.1.0

Connector keeps spamming this "Error reading next available message: invalid header" message:

{"file":"/home/jenkins/jobs/tennis-build-backend/builds/988/pkg/mod/github.com/topfreegames/pitaya@v1.1.0/service/handler.go:186","func":"github.com/topfreegames/pitaya/service.(*HandlerService).Handle","level":"error","msg":"Error reading next available message: invalid header","server":"connector","source":"tennis","time":"2020-04-22T01:06:03Z”}
Is that expected or are we doing something wrong? Apparently it happens when a user disconnects. Doesn’t seem to cause any harm, but if this is not really an unexpected error, we could make it a debug log.

**Related code:** https://github.com/topfreegames/pitaya/blob/3d3b62a19e45b2f98847424e53ae3f7ef255502a/service/handler.go#L186

felipejfc:
Does it stop logging eventually or it just keeps logging this until we kill the server?

chgz:
i remember this.
a client disconnect will get once.

my plan：
1. allocate _head_ and _body_ []byte in advance for each tcpPlayerConn  and reuse them; 
2. replace io.ReadAll with io.readFull to complete packet reading. 
@felipejfc

renatoaf:
@felipejfc it logs an error each time a client disconnects - doesn't stop

We ignored this log on filebeat

PR #150, opened 2020-12-10
Reduce connection close error logs

gabrielcorado:
When a connection closes without any message left, there is always a log that appears. This happens not even when the connection is established and closed, but in all connections being closed. This has shown to be a very annoying log (sometimes entire slots of logs are just composed by this message) and does not fully represent the error and can be misleading.

`{"component":"pitaya","level":"error","msg":"Error reading next available message: invalid header","time":"2020-12-09T21:21:25Z"}`

The idea here is to treat a connection close differently from, for example, a header with an invalid format or an incomplete message. By doing this, we are able to not log an error when it happens and reduce the noise generated in the logs.

### DIFF
--- a/acceptor/tcp_acceptor.go
+++ b/acceptor/tcp_acceptor.go
@@ -51,6 +51,10 @@ func (t *tcpPlayerConn) GetNextMessage() (b []byte, err error) {
 	if err != nil {
 		return nil, err
 	}
+	// if the header has no data, we can consider it as a closed connection
+	if len(header) == 0 {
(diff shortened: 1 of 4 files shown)


leohahn (approved):
LGTM

rodopoulos (approved):

renatoaf:
😍 this fixes https://github.com/topfreegames/pitaya/issues/138 right?

rodopoulos:
> 😍 this fixes #138 right?

Yes! :)

PR #151, opened 2020-12-10
Add more information to nats timeout error logs

victor-carvalho:
We are having a lot of problems with `nats timeout` in our clusters. We have some theories of what it might be but we need more information to make sure we are attacking the right problem. This commit adds more information to `nats timeout` logs and also add logs in case anything happens with the etcd watcher.

### DIFF
--- a/service/remote.go
+++ b/service/remote.go
@@ -98,13 +98,13 @@ func (r *RemoteService) remoteProcess(
 	switch msg.Type {
 	case message.Request:
 		if err != nil {
-			logger.Log.Errorf("Failed to process remote server with id '%s' and hostname '%s': %s", server.ID, server.Hostname, err.Error())
+			logger.Log.Errorf("Failed to process remote server with id %s and hostname %s: %s", server.ID, server.Hostname, err.Error())
 			a.AnswerWithError(ctx, msg.ID, err)
 			return
 		}
 		err := a.Session.ResponseMID(ctx, msg.ID, res.Data)
 		if err != nil {
-			logger.Log.Errorf("Failed to respond to remote server with id '%s' and hostname '%s': %s", server.ID, server.Hostname, err.Error())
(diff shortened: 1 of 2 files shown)


service/remote.go
```diff
@@ -98,13 +98,13 @@ func (r *RemoteService) remoteProcess(
...
 		if err != nil {
-			logger.Log.Errorf("Failed to process remote: %s", err.Error())
+			logger.Log.Errorf("Failed to process remote server with id '%s' and hostname '%s': %s", server.ID, server.Hostname, err.Error())
 			a.AnswerWithError(ctx, msg.ID, err)
 			return
 		}
 		err := a.Session.ResponseMID(ctx, msg.ID, res.Data)
 		if err != nil {
-			logger.Log.Errorf("Failed to respond remote: %s", err.Error())
+			logger.Log.Errorf("Failed to respond to remote server with id '%s' and hostname '%s': %s", server.ID, server.Hostname, err.Error())
```
leohahn, line 107: Just a notice, since we're using the server id on the message, it will become hard to grep for the specific message. Can't we add these as parameters of the log instead of the message?

victor-carvalho: I tried to add fields to the logger but the pitaya Log interface does not have fields. Do you know how to add them?

leohahn: Oh, so maybe there is none. I never added them. Maybe we should think about adding them to the interface them.

leohahn:
LGTM

service/remote.go
```diff
@@ -113,7 +113,7 @@ func (r *RemoteService) remoteProcess(
 			err = errors.New(res.Error.GetMsg())
 		}
 		if err != nil {
-			logger.Log.Errorf("error while sending a notify: %s", err.Error())
+			logger.Log.Errorf("error while sending a notify to server with id '%s' and hostname '%s': %s", server.ID, server.Hostname, err.Error())
```
rodopoulos, line 116: I don't think the `'` are necessary here.

rodopoulos (approved):

Issue #152, opened 2020-12-14
BeforeHandler() can't use context.WithValue()


codingcn:
BeforeHandler() should return context

PR #153, opened 2020-12-14
BeforeHandler return context

codingcn:


codingcn:
#152

codingcn:
https://github.com/topfreegames/pitaya/blob/851525f3458580c9b342efa0af9101602e79f5df/service/remote.go#L438

The code here causes the unit test to fail, which is not caused by my code.
@felipejfc

rodopoulos:
> The code here causes the unit test to fail, which is not caused by my code.

Hey, thanks for you contribution! We are addressing this failing test. We'll run the pipeline and check the PR correctness after that.

codingcn:
@rodopoulos 
Thank you for your reply. How long will it take to merge?
I think `Beforehandler` needs permission `context.WithValue ()`, we can refer to the `middleware` implementation of `gin`

felipejfc:
can you better describe why this is needed? what is the use case? I imagine that by not returning the context we are actually losing it the way it is right now?

codingcn:
> can you better describe why this is needed? what is the use case? I imagine that by not returning the context we are actually losing it the way it is right now?
@felipejfc 
For example, if I use the zap log library and add a traceid (request unique ID) to the log, I can only use it in beforehandler context.WithValue  (), but now beforehandler does not return context, which results in context.WithValue () unable to write data

codingcn:
> can you better describe why this is needed? what is the use case? I imagine that by not returning the context we are actually losing it the way it is right now?

What do you think?

codingcn:
@felipejfc Look forward to your reply.

PR #154, opened 2020-12-14
Fix tests in master

victor-carvalho:


rodopoulos (approved):

rodopoulos:
I think that's a flaky test (that we should address hehe). But try to run the CI again.

PR #143, opened 2020-09-22
Improve metrics documentation

rodopoulos:
This PR improve documentation by listing the available metrics Prometheus exports within Pitaya.

PR #157, opened 2021-01-04
Apply master (v1) latest version to v2

gabrielcorado:
The following is being ported: #139, #143, #148, #151, #150 and #154. 

The latest addition to master (#153) is not being ported in this PR. It will require some changes and will be done in a separate PR.

### DIFF
--- a/service/remote_test.go
+++ b/service/remote_test.go
@@ -473,7 +473,7 @@ func TestRemoteServiceRemoteProcess(t *testing.T) {
 			mockRPCClient.EXPECT().Call(ctx, protos.RPCType_Sys, rt, gomock.Any(), expectedMsg, gomock.Any()).Return(&protos.Response{Data: []byte("ok")}, table.remoteCallErr)
 
 			if table.remoteCallErr != nil {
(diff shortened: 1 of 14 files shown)


service/remote_test.go
```diff
@@ -68,6 +69,12 @@ func (m *MyComp) RemoteErr(ctx context.Context) (*test.SomeStruct, error) {
 
 type unregisteredStruct struct{}
 
+func errorIs(t *testing.T, err1 error, err2 error)  {
+	if !assert.ObjectsAreEqual(err1, err2) && !errors.Is(err1, err2) {
```
DylanGuedes, line 73: Shouldn't it be `!assert.ObjectsAreEqual(err1, err2) || !errors.Is(err1, err2)`?

victor-carvalho:
LGTM

victor-carvalho (approved):

rodopoulos (approved):
LGTM

service/remote_test.go
gabrielcorado: I think we assume that the struct being equal is not enough to consider it as the same (same for the opposite). But @victor-carvalho has more context on it.

DylanGuedes: Gotcha. Just double checking, anyway 😂

Issue #147, opened 2020-11-20
The import address of clientv3 needs to be updated


codingcn:
`github.com/coreos/etcd/clientv3` is out.

please use `go.etcd.io/etcd/clientv3`.

greatcat-taihe:
Please fix it. It will compile error when use new version grpc library.
../../Documents/go/pkg/mod/github.com/coreos/etcd@v3.3.9+incompatible/clientv3/health_balancer.go:54:10: undefined: grpc.Address

codingcn:
https://github.com/etcd-io/etcd/issues/12124

PR #161, opened 2021-01-13
Update dependencies

victor-carvalho:
In this PR I update most of the outdated dependencies of pitaya and also make the integration tests more easy to run on OSX (no more popup allow spam). I tried to update every dependency but it would need a lot of changes in the code because of etcd dependency that not only changed the package but also the structure of the packages.


cluster/grpc_rpc_server.go
```diff
@@ -52,8 +52,9 @@ func NewGRPCServer(config *config.Config, server *Server, metricsReporters []met
 
 // Init inits grpc rpc server
 func (gs *GRPCServer) Init() error {
+	address := gs.config.GetString("pitaya.cluster.rpc.server.grpc.address")
```
leohahn, line 55: won't address always be 0.0.0.0?

leohahn:
Does this  PR solve this issue? https://github.com/topfreegames/pitaya/issues/147

cluster/grpc_rpc_server.go
victor-carvalho: No, in the tests when starting the I pass the `-address` flag with `localhost` value.

victor-carvalho:
> Does this PR solve this issue? #147

No, to update etcd would need some changes in the code so I decided to do this in another separate PR. I don't know really know the extent of the changes as in newer versions the etcd package structure changed.

leohahn:
Fair enough, apart from the comment LGTM

rodopoulos (approved):
LGTM

cluster/grpc_rpc_server.go
leohahn: oh ,oks

leohahn (approved):

gabrielcorado (approved):

PR #182, opened 2021-04-20
conn/message: add mutex to routes/codes maps

### LINKED ISSUE #166: client panic concurrent map writes
`client.go` unsafe in high concurrency
```
routes = make(map[string]uint16) // route map to code
codes  = make(map[uint16]string) // code map to route

func SetDictionary(dict map[string]uint16) error {
    //  ....
}

```
fatal error: concurrent map writes
```
runtime.throw at panic.go:1116
runtime.mapassign_faststr at map_faststr.go:291
github.com/topfreegames/pitaya/conn/message.SetDictionary at message.go:127
github.com/topfreegames/pitaya/client.(*Client).handleHandshakeResponse at client.go:180
github.com/topfreegames/pitaya/client.(*Client).handleHandshake at client.go:404
github.com/topfreegames/pitaya/client.(*Client).ConnectTo at client.go:359
```

rochacon:
When running in high concurrency scenarios the routes/codes maps may be updated by concurrent goroutines at the same time, resulting in a race condition while calling `message.SetDictionary`. This patch adds a mutex to control access to these maps.

I decided to add the read lock to `message.GetDictionary` but this may not have the desired effect, since `map`s act as pointers [1] to its internal structures and the reference to the `routes` map is what is getting returned from the function. So, if `message.GetDictionary` callers decide to change the data in it, it may lead into another race. Ideally we would be copying the values into a new map and returning the copy. Since I'm new to the codebase I'm not sure how big this data structure can get and if this will lead into reliability issues. Should I add the map copy in there? I only found a reference to `message.GetDictionary` on `agent/agent.go:469`, where it gets serialized to JSON right after.

The new test must be called with the race detector enabled for proper validation, e.g. `go test -race ./conn/message`, however enabling it by default may result in breaking more tests than this PR is scoped to, so I decide to not enable it in the `Makefile`.

#### before the mutex addition:

```
% go test -race ./conn/message
==================
WARNING: DATA RACE
Read at 0x00c000260cf0 by goroutine 16:
  runtime.mapaccess2_faststr()
      /home/rochacon/sdk/go1.16.3/src/runtime/map_faststr.go:107 +0x0
  github.com/topfreegames/pitaya/conn/message.SetDictionary()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message.go:120 +0x184
  github.com/topfreegames/pitaya/conn/message.TestSetDictionaryRace.func1()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message_test.go:182 +0x107

Previous write at 0x00c000260cf0 by goroutine 15:
  runtime.mapassign_faststr()
      /home/rochacon/sdk/go1.16.3/src/runtime/map_faststr.go:202 +0x0
  github.com/topfreegames/pitaya/conn/message.SetDictionary()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message.go:129 +0x21a
  github.com/topfreegames/pitaya/conn/message.TestSetDictionaryRace.func1()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message_test.go:182 +0x107

Goroutine 16 (running) created at:
  github.com/topfreegames/pitaya/conn/message.TestSetDictionaryRace()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message_test.go:187 +0x13d
  testing.tRunner()
      /home/rochacon/sdk/go1.16.3/src/testing/testing.go:1193 +0x202

Goroutine 15 (finished) created at:
  github.com/topfreegames/pitaya/conn/message.TestSetDictionaryRace()
      /home/rochacon/src/github.com/topfreegames/pitaya/conn/message/message_test.go:186 +0x124
  testing.tRunner()
      /home/rochacon/sdk/go1.16.3/src/testing/testing.go:1193 +0x202
==================
--- FAIL: TestSetDictionaryRace (0.00s)
    testing.go:1092: race detected during execution of test
FAIL
FAIL    github.com/topfreegames/pitaya/conn/message     0.062s
FAIL
```

#### after

```
% go test -race ./conn/message
ok      github.com/topfreegames/pitaya/conn/message  
```

Fixes #166 

[1] https://play.golang.org/p/FKW0RnPO8J6

### DIFF
--- a/conn/message/message.go
+++ b/conn/message/message.go
@@ -139,7 +139,11 @@ func SetDictionary(dict map[string]uint16) error {
 func GetDictionary() map[string]uint16 {
 	routesCodesMutex.RLock()
 	defer routesCodesMutex.RUnlock()
-	return routes
+	dict := make(map[string]uint16)
+	for k, v := range routes {
+		dict[k] = v
+	}
+	return dict
 }
 
 func (t *Type) String() string {
--- a/conn/message/message.go
+++ b/conn/message/message.go
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 )
 
 // Type represents the type of message, which could be Request/Notify/Response/Push
@@ -54,8 +55,9 @@ var types = map[Type]string{
 }
 
 var (
-	routes = make(map[string]uint16) // route map to code
-	codes  = make(map[uint16]string) // code map to route
+	routesCodesMutex = sync.RWMutex{}
+	routes           = make(map[string]uint16) // route map to code
+	codes            = make(map[uint16]string) // code map to route
 )
 
 // Errors that could be occurred in message codec
@@ -110,6 +112,8 @@ func SetDictionary(dict map[string]uint16) error {
 	if dict == nil {
 		return nil
 	}
+	routesCodesMutex.Lock()
+	defer routesCodesMutex.Unlock()
 
 	for route, code := range dict {
 		r := strings.TrimSpace(route)
@@ -133,6 +137,8 @@ func SetDictionary(dict map[string]uint16) error {
 
 // GetDictionary gets the routes map which is used to compress route.
 func GetDictionary() map[string]uint16 {
+	routesCodesMutex.RLock()
+	defer routesCodesMutex.RUnlock()
 	return routes
 }
 
--- a/conn/message/message_encoder.go
+++ b/conn/message/message_encoder.go
@@ -69,7 +69,9 @@ func (me *MessagesEncoder) Encode(message *Message) ([]byte, error) {
 	buf := make([]byte, 0)
 	flag := byte(message.Type) << 1
 
+	routesCodesMutex.RLock()
 	code, compressed := routes[message.Route]
+	routesCodesMutex.RUnlock()
 	if compressed {
 		flag |= msgRouteCompressMask
 	}
@@ -163,7 +165,9 @@ func Decode(data []byte) (*Message, error) {
 		if flag&msgRouteCompressMask == 1 {
 			m.compressed = true
 			code := binary.BigEndian.Uint16(data[offset:(offset + 2)])
+			routesCodesMutex.RLock()
 			route, ok := codes[code]
+			routesCodesMutex.RUnlock()
 			if !ok {
 				return nil, ErrRouteInfoNotFound
 			}
--- a/conn/message/message_test.go
+++ b/conn/message/message_test.go
@@ -14,6 +14,8 @@ var update = flag.Bool("update", false, "update .golden files")
 
 func resetDicts(t *testing.T) {
 	t.Helper()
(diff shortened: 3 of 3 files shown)


felipejfc:
Hi @rochacon, thanks for the PR, it LGTM!

Regarding your question:

> Since I'm new to the codebase I'm not sure how big this data structure can get and if this will lead into reliability issues.

The maximum size of a routes map should be the amount of routes you have in total, so I would guess no more than 70 in 98% of cases...

> So, if message.GetDictionary callers decide to change the data in it, it may lead into another race. Ideally we would be copying the values into a new map and returning the copy.

I think the copy would be fine, what if we use a sync.Map? I'm trying to think about downsides of doing so (rather than breaking the API a little)

rochacon:
I rebased the branch with latest master and added the map copy in `message.GetDictionary`. I also spotted the use of the `routes` and `codes` maps in the `message_encoder.go`, so I also added the mutex calls there.

> I think the copy would be fine, what if we use a sync.Map? I'm trying to think about downsides of doing so (rather than breaking the API a little)

To be honest, I'm unsure. During my first look at the code I also thought about using a `sync.Map`. But I think the use of a mutex here is better because we keep the current type safety and avoid the need for type assertions on every use of the maps values (personally, I'm not very fond of `interface{}`). [Changes to the code](https://github.com/topfreegames/pitaya/compare/master...rochacon:conn-message-set-dict-race-with-sync-map) [would look similar](https://github.com/rochacon/pitaya/compare/conn-message-set-dict-race...rochacon:conn-message-set-dict-race-with-sync-map), it is the tests that would require more work because `assert.Equal` and `sync.Map` don't go along too well [1], we would probably need to write our own helper for that.

[1] This is the output of one of the current tests while using `sync.Map`:

```
--- FAIL: TestSetDictionary (0.00s)
    --- FAIL: TestSetDictionary/test_add_new_routes (0.00s)
        message_test.go:167:
                Error Trace:    message_test.go:167
                Error:          Not equal:
                                expected: map[string]uint16(map[string]uint16{"a":0x1, "b":0x2})
                                actual  : sync.Map(sync.Map{mu:sync.Mutex{state:0, sema:0x0}, read:atomic.Value{v:sync.readOnly{m:map[interface {}]*sync.entry{"a":(*sync.entry)(0xc00011a478)}, amended:true}}, dirty:map[interface {}]*sync.entry{"a":(*sync.entry)(0xc00011a478), "b":(*sync.entry)(0xc00011a488)}, misses:0})
                Test:           TestSetDictionary/test_add_new_routes
```

Given it is my first time messing with pitaya, I prefer not to refactor a lot of tests to make sure I'm not breaking too much stuff :D

felipejfc:
tests failed, can you take a look @rochacon ?

rochacon:
Should be working now @felipejfc, I also rebased the branch with latest master.

felipejfc:
merged! thanks for your contribution!

PR #184, opened 2021-04-26
Use single get on syncservers and parallelise parsing only

rsafonseca:


leohahn:
LGTM. Just need to check the tests why they're failing.

cluster/etcd_service_discovery.go
```diff
@@ -490,21 +491,21 @@ func (p *parallelGetter) waitAndGetResult() []*Server {
...
 }
 
 // SyncServers gets all servers from etcd
 func (sd *etcdServiceDiscovery) SyncServers() error {
-	keys, err := sd.cli.Get(
+	kvs, err := sd.cli.Get(
 		context.TODO(),
 		"servers/",
 		clientv3.WithPrefix(),
-		clientv3.WithKeysOnly(),
```
felipejfc, line 507: we better not do this every sync time. it's ok to do in server initialization though

cluster/etcd_service_discovery_test.go
```diff
@@ -493,7 +493,8 @@ func TestParallelGetter(t *testing.T) {
 
 	parallelGetter := newParallelGetter(cli, 5)
 	for _, serverToAdd := range serversToAdd {
-		parallelGetter.addWork(serverToAdd.Type, serverToAdd.ID)
+		payload := []byte("{\"id\":\"" + serverToAdd.ID + "\",\"type\":\"" + serverToAdd.Type + "\",\"frontend\":true,\"hostname\":\"" + serverToAdd.ID + "\",\"metadata\":{\"region\":\"us\"}}")
```
felipejfc, line 496: we shouldn't assume this payload, if I add a new parameter to the metadata, this will fail, right? if so then this is a no-go

felipejfc (changes requested):
Rafa, as we discussed, we shouldn't sync with values as this will (at the very minimum) add unnecessary IO overhead to ETCD.
Also, you seem to be assuming that the payload will respect a given format, which is not true, I can add whatever I want to the metadata field, or even change other fields.

cluster/etcd_service_discovery_test.go
rsafonseca: You can add more params to metadata and it won't fail, the code is just not checking for any additional parameters at this time. This dummy payload is just to ensure code coverage for the same existing functions

cluster/etcd_service_discovery.go
rsafonseca: fixed in new commit

rsafonseca:
> LGTM. Just need to check the tests why they're failing.
> 
> update: I was considering etcd-proxy usage for the PR. But @felipejfc made a good point that this may not be the case since this project is open source.

The tests themselves don't seem to be failing, the coveralls request is getting an http 500 response for some reason. Maybe expired token or something like that?

cluster/etcd_service_discovery_test.go
felipejfc: ooooh ok, I didn't see this was a test file!

felipejfc (approved):
LGTM!

PR #188, opened 2021-04-29
Fix initialisation

rsafonseca:
This fixes some potential race conditions during syncservers() execution.
- Initialises watcher before initial syncservers() run, to avoid missing put/delete events during syncservers() execution
- Watcher now waits for syncservers() to finish before processing any queued events (for example, after syncservers() gets a list of keys, if the watcher adds a new server it will then get deleted at the end of syncservers()
- Increase the default wait time from 10ms to 300ms for ETCD shutdown, since 10ms is never enough on distributed arquitectures where the servers are in different datacenters more than 10ms apart, leading to some request failures.

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -651,7 +651,7 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
 			case wResp, ok := <-chn:
 				if wResp.Err() != nil {
 					logger.Log.Warnf("etcd watcher response error: %s", wResp.Err())
-					time.Sleep(100 * time.Millisecond)
+					time.Sleep(10 * time.Millisecond)
 				}
 				if !ok {
 					logger.Log.Error("etcd watcher died, retrying to watch in 1 second")
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -37,7 +37,7 @@ import (
 )
 
 var (
-        syncServersRunning = false
+	syncServersRunning = false
 )
 
 type etcdServiceDiscovery struct {
@@ -644,7 +644,7 @@ func (sd *etcdServiceDiscovery) addServer(sv *Server) {
 
 func (sd *etcdServiceDiscovery) watchEtcdChanges() {
 	w := sd.cli.Watch(context.Background(), "servers/", clientv3.WithPrefix())
-
+	failedWatchAttempts := 0
 	go func(chn clientv3.WatchChan) {
 		for sd.running {
 			select {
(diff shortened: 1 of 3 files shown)


cluster/etcd_service_discovery.go
```diff
@@ -516,6 +520,7 @@ func (p *parallelGetter) addWork(serverType, serverID string) {
 
 // SyncServers gets all servers from etcd
 func (sd *etcdServiceDiscovery) SyncServers(firstSync bool) error {
+	syncServersRunning = true
```
felipejfc, line 523: defer func(){
   syncServersRunning = false;
}()

cluster/etcd_service_discovery.go
```diff
@@ -535,6 +540,7 @@ func (sd *etcdServiceDiscovery) SyncServers(firstSync bool) error {
 	}
 	if err != nil {
 		logger.Log.Errorf("Error querying etcd server: %s", err.Error())
+		syncServersRunning = false
```
felipejfc, line 543: then remove this one and the other one

cluster/etcd_service_discovery.go
```diff
@@ -644,14 +651,18 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
 			case wResp, ok := <-chn:
 				if wResp.Err() != nil {
 					logger.Log.Warnf("etcd watcher response error: %s", wResp.Err())
-					time.Sleep(100 * time.Millisecond)
+					time.Sleep(10 * time.Millisecond)
```
felipejfc, line 654: what is the rational for this change?

cluster/etcd_service_discovery.go
```diff
@@ -644,14 +651,18 @@ func (sd *etcdServiceDiscovery) watchEtcdChanges() {
...
 				}
 				if !ok {
 					logger.Log.Error("etcd watcher died, retrying to watch in 1 second")
 					time.Sleep(1000 * time.Millisecond)
 					sd.InitETCDClient()
 					chn = sd.cli.Watch(context.Background(), "servers/", clientv3.WithPrefix())
 				}
+				// Wait for syncServers() to finish running to avoid conflicts
+				for syncServersRunning {
```
felipejfc, line 663: instead of a for, it would be more goish and robust to use a channel or a waiting group

<- syncServersChan
or wg.Wait()

another option and probably more elegant/universal would be to use a mutex also syncServersMutex

felipejfc (changes requested):
Some little improvements to make the solution more goish

cluster/etcd_service_discovery.go
rsafonseca: I looked at the runtimes of the syncservers() in the new pitaya 1.1.8 servers we already have in production, and the function sometimes takes less than 10ms to run, so putting 100ms seemed a little excessive since we want the watcher to start processing the events again as soon as possible

rsafonseca: Actually, my mistake, this wasn't the line where the change from 100 to 10 ms was intended. fixing back to 100ms here

PR #189, opened 2021-04-30
Update lib dependencies and add some small fixes

rsafonseca:


luizmiranda7 (approved):

cluster/etcd_service_discovery.go
```diff
@@ -103,6 +106,8 @@ func (sd *etcdServiceDiscovery) configure() {
 	sd.etcdUser = sd.config.GetString("pitaya.cluster.sd.etcd.user")
 	sd.etcdPass = sd.config.GetString("pitaya.cluster.sd.etcd.pass")
 	sd.etcdDialTimeout = sd.config.GetDuration("pitaya.cluster.sd.etcd.dialtimeout")
+	sd.etcdDialKeepAliveTime = sd.config.GetDuration("pitaya.cluster.sd.etcd.dialtimeout")
+	sd.etcdDialKeepAliveTimeout = sd.config.GetDuration("pitaya.cluster.sd.etcd.dialtimeout")
```
vitorbaraujo, line 110: We should be using `pitaya.cluster.sd.etcd.dialkeepalivetime` and `pitaya.cluster.sd.etcd.dialkeepalivetimeout` here, like in `config/config.go`, right?

rsafonseca: yep, this is just adding to all locations this new configuration option, it's basically a built-in mechanism in etcd client to check if the connection is alive https://github.com/etcd-io/etcd/blob/7ea5ee8b5dcdceb5a26a5c1dfe7ca290000a007b/client/v3/config.go#L39
This isn't required for the fixes, it was just adding some extra configurability

Issue #191, opened 2021-05-06
Unhandled error with InitETCDClient(),  may panic when initializing etcd client failed


chgz:
Unhandled error with InitETCDClient() ：
https://github.com/topfreegames/pitaya/blob/master/cluster/etcd_service_discovery.go#L378
https://github.com/topfreegames/pitaya/blob/master/cluster/etcd_service_discovery.go#L665

may panic here @ https://github.com/topfreegames/pitaya/blob/master/cluster/etcd_service_discovery.go#L327
```
 panic: runtime error: invalid memory address or nil pointer dereference
 [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xdd97e9]
 goroutine 11 [running]:
 github.com/topfreegames/pitaya/cluster.(*etcdServiceDiscovery).grantLease(0xc0006c2000, 0x0, 0x0)
```

felipejfc:
fixed by #193 
thanks @chgz

PR #193, opened 2021-05-07
handle error with InitETCDClient()  #191

chgz:
[bug fix] #191 
[bug fix] handle error for tls.Listen, avoid that listener may be nil, which will cause serve() panic


felipejfc (approved):
LGTM!

Thanks for contributing!

PR #183, opened 2021-04-20
Re-add static legacy API to allow for easier migration to v2

felippeduran:
This MR addresses issue #164

PR #196, opened 2021-05-14
Add new logruswrapper constructor

felippeduran:
This MR also marks older constructors that rely on concrete implementation as deprecated.

This MR depends on #183, it's recommended that this one gets reviewed and merged afterwards.


gabrielcorado (approved):
LGTM. But we're still reviewing the #183 .

henriqueoelze:
Closing this MR assuming it has two commits on it and we split the second one into another PR (in order to unblock it)

PR #197, opened 2021-05-19
[optimize] replace uuid with nuid for requestID in processMessage()

chgz:
it's related to every message, so it benefits a lot
```
go test -bench=. -benchmem -cpu 1
goos: windows
goarch: amd64
pkg: test2
BenchmarkUUID    4627627               257 ns/op              64 B/op          2 allocs/op
BenchmarkNUID   22294016                54.0 ns/op            32 B/op          1 allocs/op
```


felipejfc:
Seems good! Can you take a look in the Unit tests? It seems they're failing.

chgz:
it's failed in e2e test ()     
case:  TestSameUIDUserShouldBeKickedInDifferentServersFromSameType
direct cause：  receiving timeout

image

i'm not sure what caused it, 
i have not the corresponding debugging environment
i think I need help about this CHECK FAILED

felipejfc:
I re-run and it worked! the e2e tests are flaky right now... merged!

PR #200, opened 2021-05-21
when processing NATS RPC requests, do not create new requests and response objects

chgz:
based on the number of goroutines of processMessages() : `pitaya.concurrency.remote.service,`

each goroutine is allocated a request and a response object in advance 

reduce memory allocation and GC expenses


felipejfc (approved):
LGTM!

Thanks @chgz

PR #204, opened 2021-06-11
[optimize] reduce response objects creation.  (submit again)

chgz:
remote call process, including **Call() , processRemoteMessage(), handleRPCSys() and handleRPCUser()** , 
the protos.Response has been created too many times.

this PR: an RPC Call can be reduced to only one response object， and use sync.pool to reuse the object.

### DIFF
--- a/service/remote.go
+++ b/service/remote.go
@@ -153,7 +153,7 @@ func (r *RemoteService) Call(ctx context.Context, req *protos.Request) (*protos.
 	}
 
 	defer tracing.FinishSpan(c, err)
-	return res, nil
+	return &*res, nil
 }
 
 // SessionBindRemote is called when a remote server binds a user session and want us to acknowledge it
--- a/service/remote.go
+++ b/service/remote.go
@@ -25,6 +25,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 
 	"github.com/golang/protobuf/proto"
 
@@ -58,6 +59,7 @@ type RemoteService struct {
 	messageEncoder         message.Encoder
 	server                 *cluster.Server // server obj
 	remoteBindingListeners []cluster.RemoteBindingListener
+	responsePool           *sync.Pool
 }
 
 // NewRemoteService creates and return a new RemoteService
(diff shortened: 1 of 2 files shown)


service/remote.go
```diff
@@ -136,15 +145,15 @@ func (r *RemoteService) Call(ctx context.Context, req *protos.Request) (*protos.
...
+		processRemoteMessage(c, req, res, r)
 	}
 
 	if res.Error != nil {
 		err = errors.New(res.Error.Msg)
 	}
 
 	defer tracing.FinishSpan(c, err)
-	return res, nil
+	return &*res, nil
```
felipejfc, line 156: why not simply return res instead of &*res?

service/remote.go
```diff
@@ -127,7 +134,9 @@ func (r *RemoteService) AddRemoteBindingListener(bindingListener cluster.RemoteB
 func (r *RemoteService) Call(ctx context.Context, req *protos.Request) (*protos.Response, error) {
 	c, err := util.GetContextFromRequest(req, r.server.ID)
 	c = util.StartSpanFromRequest(c, r.server.ID, req.GetMsg().GetRoute())
-	var res *protos.Response
+	res := r.responsePool.Get().(*protos.Response)
+	res.Reset()
+	defer r.responsePool.Put(res)
```
felipejfc, line 139: given that this Pool will be used by many threads, don't we have a racing condition here, where other thread gets the same "res" object that is being used by the call stack after the return of this function?

chgz: > A Pool is safe for use by multiple goroutines simultaneously.
sync.Pool's  Get() and Put()  are both concurrently safe.

so，we don't got data race here.

service/remote.go
chgz: Because the return value of Call() is a pointer.
And, we want to put it back in the pool and reuse it.
Therefore, we can only return its copy value here to avoid object insecurity.

service/remote.go
chgz: Sorry for did not reply in time!
because it is the Dragon Boat Festival holiday in my country these few days.

service/remote.go
felipejfc: Oh, I see!

service/remote.go
felipejfc: no problem!

felipejfc:
LGTM! Thanks for the contribution @chgz

PR #208, opened 2021-06-18
fix(pomelo_packet_decoder): fix read buff

lx-world:


felipejfc (changes requested):
I think these changes are not thread-safe, we reutilize the same packet decoder in all connections. Also I still don't understand what problem you're trying to solve here. Can you give an example?

lx-world:
> I think these changes are not thread-safe, we reutilize the same packet decoder in all connections. Also I still don't understand what problem you're trying to solve here. Can you give an example?


If there are three and a half packs in the ***buf**, what should I do?

henriqueoelze:
We talked internally and understood that it didn't make sense to merge it into master now. If you believe that this is crucial to you, you can open an Issue and we could discuss the solution there in the future.

PR #210, opened 2021-06-21
some logic optimiztaion

chgz:
- prometheus reporter startup with panic
      **when** we start multiple pitaya services on the same host, it will panic because the port of prometheus reporter monitoring fails. Often occurs in the test environment.

- err of remote call should be added with routing information
      **Like**, `msg="error making call to target with id dd953bc8-efbc-4e55-a95a-7ab794a8bcc3 and host xxxx: %!w(*errors.Error=&{PIT-000 xxxxxx map[]})"`  .  I think it is useful to show which handler is called.

- when serverID is empty, execute remoteCall directly. Avoid executing sd.GetServer() once
  logic optimiztaion

### DIFF
--- a/metrics/prometheus_reporter.go
+++ b/metrics/prometheus_reporter.go
@@ -22,7 +22,7 @@ package metrics
 
 import (
 	"fmt"
-	"log"
+	"github.com/topfreegames/pitaya/logger"
 	"net/http"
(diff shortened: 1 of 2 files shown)


henriqueoelze (approved):
LGTM

PR #225, opened 2021-07-28
[bugfix] Protoclient was crashing if any handler had no outputs

gabrielcorado:
The problem was related to handlers with no output since the `getOutputInputNames` directly iterated into the outputs without checking any.

Besides the fix, some other changes were made:
* Updated the errors in the `protoclient` connection flow to be more descriptive;
* Updated the `examples/demo/cluster` to only used Protobuf. (This example was used to validate the fix since it has the `Message` handler, which doesn't have any output).


henriqueoelze (approved):
🚢

PR #230, opened 2021-08-06
Change initial size of get servers return map

### LINKED ISSUE #229: There's something wrong with this function: etcdServiceDiscovery.GetServersByType()
```
ret := make(map[string]*Server, len(sd.serverMapByType))
for k, v := range sd.serverMapByType[serverType] {
	ret[k] = v
}
```
The cap value of make is not very reasonable

gabrielcorado:
Closes #229 

The initial size of the map used to return the servers (on ETCD service discovery) was using an incorrect reference. Even though this mistake didn't cause any issue, changing it makes it less confusing while reading the code.

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -304,7 +304,7 @@ func (sd *etcdServiceDiscovery) GetServersByType(serverType string) (map[string]
 		// Create a new map to avoid concurrent read and write access to the
 		// map, this also prevents accidental changes to the list of servers
 		// kept by the service discovery.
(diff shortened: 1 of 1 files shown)


henriqueoelze (approved):
LGTM, but do we have some unit test that we can change to validate it?

gabrielcorado:
Not really, since this isn't a bug fix, only a change on the code to make it more readable. We could have a map with an initial size less than necessary, but since it is the initial size, there is no problem with adding more elements. (here is a short example: https://play.golang.org/p/RxcuCNQ5lNG)

PR #278, opened 2022-07-12
Fix concurrency bottleneck, improve shutdown draining

rsafonseca:
Fixes a bottleneck to concurrent processing caused by the handleMessages() function by offloading most of the processing handled by this function to the processMessage() workers, and avoiding an additional data copy to another channel.


felipejfc (changes requested):
Nice work Rafa! Can we remove from the init method the goroutine with shutdown logic? Also, can you share some benchmarks comparing the old code with this new one?

PR #279, opened 2022-07-18
Add SimpleEvent() method to Reporter interface

gibertoni:
Add SimpleEvent method to Reporter interface and implement it on StatsD reporter. 
This shall add an event entry in the configured namespace and be used to track important changes into deployments and services.


metrics/statsd_reporter.go
```diff
@@ -33,6 +33,7 @@ import (
 type Client interface {
 	Count(name string, value int64, tags []string, rate float64) error
 	Gauge(name string, value float64, tags []string, rate float64) error
+	SimpleEvent(title string, text string) error
```
adilsonwild, line 36: Why `SimpleEvent` and not just `Event`? Are there other types of events?

metrics/reporter_interfaces.go
```diff
@@ -25,4 +25,5 @@ type Reporter interface {
 	ReportCount(metric string, tags map[string]string, count float64) error
 	ReportSummary(metric string, tags map[string]string, value float64) error
 	ReportGauge(metric string, tags map[string]string, value float64) error
+	ReportEvent(title string, text string) error
```
adilsonwild, line 28: The other methods also accept `tags`. Do events support tags? If so, we may add them to this method as well.

metrics/prometheus_reporter.go
```diff
@@ -344,6 +344,11 @@ func (p *PrometheusReporter) ReportGauge(metric string, labels map[string]string
 	return constants.ErrMetricNotKnown
 }
 
+// ReportEvent is not implemented by PrometheusReporter
+func (p *PrometheusReporter) ReportEvent(title string, text string) error {
+	return nil
```
adilsonwild, line 349: Since Prometheus doesn't provide a way to report events, we should at least log when someone is trying to use this method (throw a warning or even an error log).

adilsonwild: Actually, returning an error explaining the feature is not supported might be even better 🤔

metrics/statsd_reporter.go
gibertoni: It matches the interface from datadog-go package: https://github.com/DataDog/datadog-go/blob/master/statsd/statsd.go#L191

metrics/reporter_interfaces.go
gibertoni: As far as I looked events do not support tags

metrics/prometheus_reporter.go
gibertoni: Agreed! Pushed the change to return an error.

metrics/prometheus_reporter.go
```diff
@@ -344,6 +345,11 @@ func (p *PrometheusReporter) ReportGauge(metric string, labels map[string]string
 	return constants.ErrMetricNotKnown
 }
 
+// ReportEvent is not implemented by PrometheusReporter
+func (p *PrometheusReporter) ReportEvent(title string, text string) error {
+	return xerrors.Errorf("Event reporting is not supported by PrometheusReporter")
```
felipejfc, line 350: For not implemented methods, return the error that's defined in the constants package.
e.g.
https://github.com/topfreegames/pitaya/blob/2829a35ad96b87812b2b3041c4b40fb3576fe1ce/cluster/grpc_rpc_client.go#L153-L155

metrics/prometheus_reporter.go
```diff
@@ -23,6 +23,7 @@ package metrics
 import (
 	"fmt"
 	"github.com/topfreegames/pitaya/logger"
+	"golang.org/x/xerrors"
```
felipejfc, line 26: don't think you need this dependency. you should use constants

metrics/statsd_reporter_test.go
```diff
@@ -23,6 +23,7 @@ package metrics
 import (
 	"errors"
 	"fmt"
+	"golang.org/x/xerrors"
```
felipejfc, line 26: Same as other comments

felipejfc (changes requested):
The implementation LGTM. Only have some request regarding how you're dealing with errors to be compliant with the rest of the project.

metrics/statsd_reporter_test.go
gibertoni: Fixed in latest commit

metrics/prometheus_reporter.go
gibertoni: Fixed in latest commit

metrics/prometheus_reporter.go
gibertoni: Fixed in latest commit

felipejfc (approved):

PR #280, opened 2022-07-25
Fix all test flakes, fix compat with opentelemetry 1.x

rsafonseca:
This MR fixes the following:
- Upgrade ETCD to 3.6.0-alpha.0 to resolve problems with OpenTelemetry imports (this caused problems in the testing package when using other modules that relied on OpenTelemetry 1.x, eg. wildlife-modules)
- Fix flaky e2e tests
- Fix flaky ws_acceptor test
- Fix test leaks in group_service_test
- Upgrade mock from 1.4.4 to 1.6.0
- Bump go version to 1.17
- Upgrade NATS from 2.2.0 to 2.8.4 to fix dependabot critical security vulnerabilities alerts


cluster/nats_rpc_server_test.go
```diff
@@ -335,7 +335,7 @@ func TestNatsRPCServerInitShouldFailIfConnFails(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	mockSessionPool := sessionmocks.NewMockSessionPool(ctrl)
 	rpcServer, _ := NewNatsRPCServer(*cfg, sv, nil, nil, mockSessionPool)
-	mockSessionPool.EXPECT().OnSessionBind(rpcServer.onSessionBind)
+	//mockSessionPool.EXPECT().OnSessionBind(rpcServer.onSessionBind)
```
felipejfc, line 338: do you know why this is not happening anymore?

rsafonseca: I think this was an assertion bug on older mock, this is a duplicate expect, you'll see the other one a few lines below.

felipejfc (approved):

PR #284, opened 2022-08-15
Add acceptor support for receiving pre-loaded certificates

felippeduran:
This MR exposes a new constructor for the TCP acceptor that allows for pre-loaded certificates as arguments.

It also refactors internal code in order to make sure the new constructor is used by the old one and make appropriate changes to store a list of pre-loaded certificates instead of a `certFile`/`certKey` pair.

Tests were changed accordingly.

You could say that there's a tiny API change, which is that the `NewTCPAcceptor` might now panic if the provided certificate file paths are invalid, but it doesn't seem to me as something that relevant as its actually related to a failed initialization case where pitaya would panic somewhere else anyway (inside `ListenAndServeTLS`).


acceptor/tcp_acceptor.go
```diff
@@ -78,21 +77,26 @@ func (t *tcpPlayerConn) GetNextMessage() (b []byte, err error) {
...
+		if err != nil {
+			panic(constants.ErrInvalidCertificates)
+		}
+		certificates = append(certificates, cert)
 	}
 
+	return NewTCP(addr, certificates...)
+}
+
+func NewTCP(addr string, certs ...tls.Certificate) *TCPAcceptor {
```
vitorbaraujo, line 94: Given that `ListenAndServeTLS` requires providing a certificate and key, if you want to use this constructor, you would need to provide the same certificate information here and in the listen method? In more general terms, I cannot configure all certificates here and then just call the listen method without duplicating some of the certificates information, right?

acceptor/tcp_acceptor.go
```diff
@@ -78,21 +77,26 @@ func (t *tcpPlayerConn) GetNextMessage() (b []byte, err error) {
...
-	certFile := ""
+	certificates := []tls.Certificate{}
 	if len(certs) != 2 && len(certs) != 0 {
 		panic(constants.ErrInvalidCertificates)
 	} else if len(certs) == 2 {
-		certFile = certs[0]
-		keyFile = certs[1]
+		cert, err := tls.LoadX509KeyPair(certs[0], certs[1])
+		if err != nil {
+			panic(constants.ErrInvalidCertificates)
```
vitorbaraujo, line 86: It might be good to add a test validating this behavior.

acceptor/tcp_acceptor.go
vitorbaraujo: This behavior seems to be the same before the changes as well, to be honest.

reinaldooli: I don't know exactly why we have this one, maybe we should add a similar method that receives the certs. Anyway, the `ListAndServe` method already check for the certificates given through the `New...` methods.

acceptor/tcp_acceptor.go
reinaldooli: There are some already: https://github.com/topfreegames/pitaya/pull/284/files#diff-91d8960c4bbf5e4ec59f1cc6232c15ea612244fef31e5a2e8c49edbd4c348997L52

vitorbaraujo: These current tests only seem to assert the case when it panics when `len(certs)` is 0 or different than 2. I believe we need to add a new entry to the test table with two invalid entries in `certs` so it tries to load the certificate and then panics.

acceptor/tcp_acceptor.go
vitorbaraujo: Oh, I see, the `ListenAndServe` will fall back to the TLS variant if there are any certificates provided in the constructor. LGTM, then.

vitorbaraujo (approved):

acceptor/tcp_acceptor.go
```diff
@@ -117,13 +121,13 @@ func (a *TCPAcceptor) Stop() {
 }
 
 func (a *TCPAcceptor) hasTLSCertificates() bool {
-	return a.certFile != "" && a.keyFile != ""
+	return len(a.certs) > 0
```
rsafonseca, line 124: Shouldn't this be != 2 ? Otherwise it will return success if it only has 1 cert or if it has 10

acceptor/tcp_acceptor.go
rsafonseca, line 94: This name seems fairly generic, could we name it something like NewTLSAcceptor something like that?

felippeduran: That's a good question, I always struggle with naming in Golang.

I defined it as `NewTCP` so that it doesn't stutter when we have it on code like `acceptor.NewTCP` (instead of `acceptor.NewTCPAcceptor`). The more verbose option would also conflict with the previous existing definition.

Another option could be `NewTCPWithCertificates`.

What do you think?

acceptor/tcp_acceptor.go
felippeduran: If I recall correctly, I think that the TLS API supports multiple certificates. So in order to make the "new" tcp acceptor API fully functional, I left it as a more generic checking.

The constraint for 2 files is only related to the "legacy" API that loads the files itself. I think we can remove this altogether in the future, making the code much more generic.

What do you think?

acceptor/tcp_acceptor.go
felipejfc: Given the current naming standards, I agree with Rafa on changing to NewTLSAcceptor

felipejfc (approved):

PR #295, opened 2023-01-19
Fix some connector rpc messaging issues, avoid ETCD log noise when using etcd-proxy

rsafonseca:


service/remote.go
```diff
@@ -140,24 +141,46 @@ func (r *RemoteService) AddRemoteBindingListener(bindingListener cluster.RemoteB
...
-		res = processRemoteMessage(c, req, r)
 	}
 
 	if res.Error != nil {
 		err = errors.New(res.Error.Msg)
 	}
 
-	defer tracing.FinishSpan(c, err)
-	return res, nil
+	return res, err
```
rsafonseca, line 183: reminder: this error was always ignored, especially if the error is timeout related, it's important that at least the server logs it. this is also being discarded in the caller function and just passes the error to the response sent to the caller, but the caller might not get that message (it might have already gone away or timed out waiting)

service/remote.go
```diff
@@ -140,24 +141,46 @@ func (r *RemoteService) AddRemoteBindingListener(bindingListener cluster.RemoteB
...
 	var res *protos.Response
+
+	if err == nil {
+		result := make(chan *protos.Response, 1)
+		go func() {
+			result <- processRemoteMessage(c, req, r)
+		}()
+
+		// Set a timeout for processing the call
+		timeout := time.Duration(300) * time.Second
```
felipejfc, line 154: Why 300 seconds? How did you come up with this number? Lets make it configurable and use a smaller time :)

service/remote.go
```diff
@@ -140,24 +141,46 @@ func (r *RemoteService) AddRemoteBindingListener(bindingListener cluster.RemoteB
...
+		if ctx.Value("reqTimeout") != nil {
+			timeout, err = time.ParseDuration(fmt.Sprint(ctx.Value("reqTimeout")))
+			if err != nil {
+				logger.Log.Errorf("Error while parsing timeout duration: %s", err.Error())
+			}
+		}
+
+		select {
+		case <-time.After(timeout):
+			err = fmt.Errorf("Timed out calling route %s after %d seconds.", req.GetMsg().GetRoute(), timeout)
```
felipejfc, line 164: In this case, will the request be cancelled or will it keep hanging until something happens? I was expecting to see a context.Cancell or something else

session/session.go
```diff
@@ -92,6 +92,7 @@ type sessionImpl struct {
 	frontendID        string                      // the id of the frontend that owns the session
 	frontendSessionID int64                       // the id of the session on the frontend server
 	Subscriptions     []*nats.Subscription        // subscription created on bind when using nats rpc server
+	requestsInFlight  map[string]string           // whether the session is waiting from a response from a remote
```
felipejfc, line 95: Will this be consumed by multiple goroutines, in which case we should use sync.Map?

cluster/nats_rpc_client.go
```diff
@@ -197,6 +212,7 @@ func (ns *NatsRPCClient) Call(
 		}
 		return nil, err
 	}
+	logger.Log.Debug("returning response from remote (client)")
```
felipejfc, line 215: Why do we need this log line, even being a debug it seems to be useless

cluster/nats_rpc_client.go
```diff
@@ -155,7 +156,21 @@ func (ns *NatsRPCClient) Call(
...
+		requestInfo := ""
+		if route != nil {
+			requestInfo = route.Method
+		}
+
+		session.SetRequestInFlight(requestID, requestInfo, true)
+		defer session.SetRequestInFlight(requestID, "", false)
+	}
+
+	logger.Log.Debugf("sending remote nats request (client) with timeout of %s", ns.connectionTimeout)
```
felipejfc, line 171: do we need this log line? we already have log for sending requests

felipejfc (changes requested):
Left some comments. Next time try to split into multiple PRs instead of a single with multiple changes :)

acceptor/acceptor.go
```diff
@@ -34,4 +35,5 @@ type Acceptor interface {
 	Stop()
 	GetAddr() string
 	GetConnChan() chan PlayerConn
+	EnableProxyProtocol()
```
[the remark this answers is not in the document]
rsafonseca: sadly it doesn't work. once you read from the socket, if the header is missing, you will be missing data on the next calls

cluster/etcd_service_discovery.go
```diff
@@ -339,9 +340,11 @@ func (sd *etcdServiceDiscovery) InitETCDClient() error {
 	logger.Log.Infof("Initializing ETCD client")
 	var cli *clientv3.Client
 	var err error
+	etcdClientLogger, _ := logutil.CreateDefaultZapLogger(logutil.ConvertToZapLevel("error"))
```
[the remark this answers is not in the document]
rsafonseca: because this doesn't use our internal logger. This is to tap into the etcd client library's own logger and set it to a different log level than our internal logger (we don't want the noisy warn+info messages from the etcd library, but we do want the warn/info messages which are controlled by the log level settings on our app).

cluster/nats_rpc_client.go
rsafonseca: we don't, this was accidentally left during testing, thanks for the catch, i'll remove it

cluster/nats_rpc_client.go
rsafonseca: we don't, same as above, left from debug testing.. removing it

service/remote.go
rsafonseca: This is just a safety net to ensure there is a set value (in case backend is updated and frontend is not), this is also the existing hardcoded default on the properties below. This is already configurable, the developer just needs to set the request timeout with "pitaya.cluster.rpc.client.nats.requesttimeout" or "pitaya.cluster.rpc.client.grpc.requesttimeout".

rsafonseca: Actually the hardcoded default is 5 seconds, not 5 minutes 😂 in any case, the default on the backend might not be the same as the default on the frontend, hence a higher default is safer. Here's an example:

service/remote.go
rsafonseca: Currently, the function called by the route will keep running, but the processing slot will be freed up preventing the queue to fill. Ideally you're right, it should cancel the context as well, just to ensure resources are freed up immediately. I'll add the required changes :)

cluster/nats_rpc_client.go
rsafonseca: Apparently we don't. This was useful for me to figure out which part of the request things were hanging up on, although the line isn't very insightful (the route for example and some request ID would be helpful on this log)

rsafonseca: image
Here's an example log with debug enabled.. there's nothing logging the request

cluster/etcd_service_discovery.go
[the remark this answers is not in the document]
felipejfc: Does ETCD support replacing its internal logger with an instance of ours? Like implementing an interface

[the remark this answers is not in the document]
rsafonseca: It doesn't and even if it did, the noisy warning message we want to get rid of is also not something which we can set to debug. So we'll need to have two loggers with different levels (the app one that the user can control, and the etcd one, set to error level explicitly)

acceptor/acceptor.go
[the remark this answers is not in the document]
rsafonseca: Not without overhauling a lot of things. It's easier to just configure this as a flag, which is what all softwares that I know do, e.g. nginx

service/remote.go
rsafonseca: addressed

service/remote.go
rsafonseca: Actually, upon some real life testing, looks like the timeout isn't making its way to the server (metadata isn't being unmarshalled) need to fix this, hold on ;)

session/session.go
rsafonseca: Added a mutex instead, simplifies counting the values and should be slightly more performant :)

session/session.go
```diff
@@ -92,10 +92,15 @@ type sessionImpl struct {
 	frontendID        string                      // the id of the frontend that owns the session
 	frontendSessionID int64                       // the id of the session on the frontend server
 	Subscriptions     []*nats.Subscription        // subscription created on bind when using nats rpc server
-	requestsInFlight  map[string]string           // whether the session is waiting from a response from a remote
+	requestsInFlight  ReqInFlight           // whether the session is waiting from a response from a remote
 	pool              *sessionPoolImpl
 }
 
+type ReqInFlight struct {
```
felipejfc, line 99: why not use sync.Map?

PR #308, opened 2023-05-12
Add handshake validators

reinaldooli:
## Context

Sometimes, it is mandatory that the clients met some requirements before being allowed to connect.<br />
In order to achieve this, `SessionPool` now has a function `AddHandshakeValidator` and all the added validators will run as soon as the handshake is processed. <br>
With this, clients can send the requirements through handshake parameters, and it can be validated before establishing the connection.


service/handler.go
```diff
@@ -227,6 +227,13 @@ func (h *HandlerService) processPacket(a agent.Agent, p *packet.Packet) error {
 			return fmt.Errorf("Invalid handshake data. Id=%d", a.GetSession().ID())
 		}
 
+		for _, fun := range a.GetSession().GetHandshakeValidators() {
+			if err := fun(handshakeData); err != nil {
+				a.SetStatus(constants.StatusClosed)
+				return fmt.Errorf("Handshake validation failed: %w. Id=%d", err, a.GetSession().ID())
```
felippeduran, line 233: Is it possible to return specific error codes to clients if the validation fail? That'd be very useful to handle forced updates and server maintenance, for example.

service/handler.go
```diff
@@ -227,6 +227,13 @@ func (h *HandlerService) processPacket(a agent.Agent, p *packet.Packet) error {
 			return fmt.Errorf("Invalid handshake data. Id=%d", a.GetSession().ID())
 		}
 
+		for _, fun := range a.GetSession().GetHandshakeValidators() {
+			if err := fun(handshakeData); err != nil {
```
vitorbaraujo, line 231: The way it is, an error is returned when the first validation fails, ignoring following validations. Would it be interesting to run all validators and return an error if any one of them fails? This would provide less back and forth in case multiple validations are to fail.

vitorbaraujo: To be honest, I'm not sure if this use-case if very valuable, though.

felipejfc: For simplicity sake's I say lets keep the way it is right now. The end result is the same whether multiple or only one validator failed.

service/handler.go
felipejfc: Isn't the err here already the one returned by the handshake validator? And if so it is something you have control over already, right?

felipejfc (commented):
Did you test with no handshake validators added? I missed where you are initializing the array

service/handler.go
reinaldooli: In this step, clients already received a message about the [handshake](https://github.com/topfreegames/pitaya/blob/feature/handshake-validators/service/handler.go#L216).
Any error that may occur after sending the handshake response, just close the session.

felipejfc:
@reinaldooli please update the docs with HandshakeValidators information

service/handler.go
felippeduran: Hum, I see. I guess it'd be best to return the handshake response only after validation.

From the client perspective, getting a sudden connection close without reason might be very confusing during development, wouldn't it?

felippeduran (commented):
Other the the comment I raised, the rest looks good. I also agree with the conclusion in @vitorbaraujo's thread.

However, I believe it'd be important to resolve my thread before merging.

service/handler.go
reinaldooli: Yes! I've changed the process to send more comprehensive messages to clients. This first version should not break anything on the client side, but it won't send any information about the error to clients, just a 400 code (Bad Request). A further (maybe breaking) change can be addressed to provide to client a clearer message about what happened in the server.

PR #331, opened 2023-08-02
Rename the key according to doc/configuration.rst line 78 and 275

sandy1219:


sandy1219:
* - pitaya.cluster.sd.etcd.syncservers.parallelism
    - 10
    - int
    - The number of goroutines that should be used while getting server information on etcd initialization
 
  * - pitaya.metrics.period
    - 15s
    - string
    - Period that system metrics will be reported

felipejfc (approved):

PR #335, opened 2023-08-11
Fixed an issue where variables were printed with the wrong placeholder

haoyu234:
Here are some sample outputs:

before
```
DEBU[0000] adding server &{f8da4226-e555-d75f-bf35-e7e3f9e84945 chat map[] %!s(bool=true) h.dev}  source=pitaya
DEBU[0000] subs channel size: %!d(float64=0), max: %!d(float64=0), dropped: 0  source=tank
```

after
```
DEBU[0000] adding server &{5589a691-ad56-4d31-bca2-55d742d97ece connector map[] true h.dev}  source=pitaya
DEBU[0000] subs channel size: 0, max: 0, dropped: 0      source=pitaya
```

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -578,7 +578,7 @@ func (sd *etcdServiceDiscovery) SyncServers(firstSync bool) error {
 	servers := parallelGetter.waitAndGetResult()
 
 	for _, server := range servers {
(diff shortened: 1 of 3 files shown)


felipejfc (approved):

felipejfc:
thanks for the contribution

PR #337, opened 2023-08-29
Fix stale ETCD routines when ETCD server is unavailable

rsafonseca:
- Ensures DialTimeout is respected (this has been broken since it became async on etcd client 3.3.x)
- Sets a deadline on Lease Grant call to ensure it doesn't block forever
- Adds error logs when nats client detects a slow consumer
- Reduces the noise of logs regarding failure to read messages when the remote client has already closed the connection

### DIFF
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -188,7 +188,7 @@ func (sd *etcdServiceDiscovery) renewLease() error {
 
 func (sd *etcdServiceDiscovery) grantLease() error {
 	// grab lease
-	ctx, cancel := context.WithTimeout(context.Background(), sd.heartbeatTTL)
+	ctx, cancel := context.WithTimeout(context.Background(), sd.etcdDialTimeout)
 	defer cancel()
 	l, err := sd.cli.Grant(ctx, int64(sd.heartbeatTTL.Seconds()))
 	if err != nil {
--- a/cluster/etcd_service_discovery.go
+++ b/cluster/etcd_service_discovery.go
@@ -188,7 +188,7 @@ func (sd *etcdServiceDiscovery) renewLease() error {
 
 func (sd *etcdServiceDiscovery) grantLease() error {
 	// grab lease
-	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(sd.heartbeatTTL.Seconds()))
+	ctx, cancel := context.WithTimeout(context.Background(), sd.heartbeatTTL)
 	defer cancel()
 	l, err := sd.cli.Grant(ctx, int64(sd.heartbeatTTL.Seconds()))
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cluster/etcd_service_discovery.go
```diff
@@ -187,7 +188,9 @@ func (sd *etcdServiceDiscovery) renewLease() error {
 
 func (sd *etcdServiceDiscovery) grantLease() error {
 	// grab lease
-	l, err := sd.cli.Grant(context.TODO(), int64(sd.heartbeatTTL.Seconds()))
+	ctx, cancel := context.WithTimeout(context.Background(), sd.heartbeatTTL)
```
felipejfc, line 191: create a new option for acquire lease timeout.

felipejfc (changes requested):
Lets create a new option for the timeout and we're good to go

cluster/etcd_service_discovery.go
rsafonseca: Isn't the heartbeatTTL a good value for the timeout?

service/handler.go
```diff
@@ -181,7 +183,12 @@ func (h *HandlerService) Handle(conn acceptor.PlayerConn) {
 		msg, err := conn.GetNextMessage()
 
 		if err != nil {
-			if err != constants.ErrConnectionClosed {
+			// Check if this is an expected error due to connection being closed
+			if errors.Is(err, net.ErrClosed) {
+				logger.Log.Debugf("Connection no longer available while reading next available message: %s", err.Error())
+			} else if err == constants.ErrConnectionClosed {
+				logger.Log.Debugf("Connection no longer available while reading next available message: %s", err.Error())
```
vitorbaraujo, line 190: Are both logs supposed to be equal?

cluster/nats_rpc_common.go
```diff
@@ -52,6 +52,15 @@ func setupNatsConn(connectString string, appDieChan chan bool, options ...nats.O
 				appDieChan <- true
 			}
 		}),
+		nats.ErrorHandler(func(nc *nats.Conn, sub *nats.Subscription, err error) {
+			if err == nats.ErrSlowConsumer {
+				dropped, _ := sub.Dropped()
+				logger.Log.Warn("Slow consumer on subject %q: dropped %d messages\n",
+					sub.Subject, dropped)
+			} else {
+				logger.Log.Errorf(err.Error())
```
vitorbaraujo, line 61: It might be good to mention nats in the logs for easy readability.

cluster/etcd_service_discovery.go
felipejfc: I believe it's not. The heartbeatTTL is something the game has control over and can increase or decrease based on how permissive they want to be with unstable connections. This one on the other hand is something internal that we can come up with a sweet spot value

rsafonseca: Well, shouldn't that sweet spot be about the same as the ETCD dial timeout? Maybe we can use the same since ultimately it's for a similar use: the deadline of a GRPC call to ETCD server

felipejfc: makes sense, we can go with it then

rsafonseca: done

felipejfc (approved):

Issue #289, opened 2022-10-22
bug: connect not close when Handshake failed


lolizeppelin:
err := json.Unmarshal(p.Data, handshakeData)
if err != nil {
    a.SetStatus(constants.StatusClosed)

where  Unmarshal failed, StatusClosed will be set 

then agent.Close() will find  a.GetStatus() == constants.StatusClosed

so  sokcet will not be close

PR #343, opened 2023-09-13
Fix agent socket cleanup

rsafonseca:
Fixes https://github.com/topfreegames/pitaya/issues/289
Resolves several test flakes
Ensures Acceptors are listening before moving forward with initialization

### DIFF
--- a/acceptor/ws_acceptor.go
+++ b/acceptor/ws_acceptor.go
@@ -41,6 +41,7 @@ type WSAcceptor struct {
 	listener net.Listener
 	certFile string
 	keyFile  string
+	running  bool
 }
 
 // NewWSAcceptor returns a new instance of WSAcceptor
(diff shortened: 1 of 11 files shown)


acceptor/ws_acceptor.go
```diff
@@ -160,6 +170,7 @@ func (w *WSAcceptor) serve(upgrader *websocket.Upgrader) {
 
 // Stop stops the acceptor
 func (w *WSAcceptor) Stop() {
+	w.running = false
```
felipejfc, line 173: only if the below lines didn't error

felipejfc (changes requested):

acceptor/ws_acceptor.go
rsafonseca: once you issue the Stop() it should be marked as not running anyway, since there might be an error where it couldn't run close() because it was already closed or other issue, and in that situation, the acceptor would keep getting marked as running, when it really wasn't (if it can't even close itself, it should not have running=true). This matches the existing implementation on the TCPAcceptor

app.go
```diff
@@ -366,6 +366,10 @@ func (app *App) listen() {
 			a.ListenAndServe()
 		}()
 
+		for a.IsRunning() == false {
+			logger.Log.Infof("Waiting for TCP acceptor %s to start on addr %s", reflect.TypeOf(a), a.GetConfiguredAddress())
+		}
+
 		logger.Log.Infof("listening with acceptor %s on addr %s", reflect.TypeOf(a), a.GetAddr())
```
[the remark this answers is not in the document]
felipejfc: I'm pretty sure this log line is broken though, it's not properly printing the port

felipejfc (approved):

PR #345, opened 2023-09-13
Fix cli install

rsafonseca:
- The documented cli install commands aren't working and there's no simple one-line command to install this without cloning the repo, since there's a mismatch between the module name and its path.
- Bump pitaya version on cli dependencies


felippeduran (approved):
Looks good to me! As long as CI was already failing before, seems fine to merge it.

rsafonseca:
> Looks good to me! As long as CI was already failing before, seems fine to merge it.

It was not, fixing that now

PR #351, opened 2023-10-09
Support before/after hooks on user remote RPCs

renatoaf:
Currently, "before" / "after" hooks are invoked only on "sys" RPCs on `remote.go` module.

Therefore there's no way to add before/after hooks on "user" remote requests, only via handlers.

This MR invokes the existing hooks on remote pipeline with `ExecuteBeforePipeline` and `ExecuteAfterPipeline`.

### DIFF
--- a/service/remote.go
+++ b/service/remote.go
@@ -334,8 +334,6 @@ func processRemoteMessage(ctx context.Context, req *protos.Request, r *RemoteSer
 }
 
 func (r *RemoteService) handleRPCUser(ctx context.Context, req *protos.Request, rt *route.Route) *protos.Response {
-	response := &protos.Response{}
-
 	remote, ok := r.remotes[rt.Short()]
 	if !ok {
 		logger.Log.Warnf("pitaya/remote: %s not found", rt.Short())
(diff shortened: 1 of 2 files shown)


felipejfc (commented):
I think we should create another set of hooks "remoteHooks" for Remotes as they're very different than Handlers.
Remotes -> General Server to Server RPC
Handlers -> Frontend to Backend RPC generated from a player handler call.

renatoaf:
I considered doing that initially also to avoid introducing this behavior change unexpectedly, however I figured that every kind of RPC share the same "HandlerHooks" struct. For example "sys.push" and "sys.bindsession" already trigger the same set of HandlerHooks we have. Should be a small change, do you think it's worth doing the separation just for remotes?

felipejfc:
> I considered doing that initially also to avoid introducing this behavior change unexpectedly, however I figured that every kind of RPC share the same "HandlerHooks" struct. For example "sys.push" and "sys.bindsession" already trigger the same set of HandlerHooks we have. Should be a small change, do you think it's worth doing the separation just for remotes?

I believe it is, so that we reduce the probability of human error when creating these hooks. The separation will help with clarity of what will run when. Can you adjust the PR?

renatoaf:
Closing this as the work was finished on this new PR: https://github.com/topfreegames/pitaya/pull/365

PR #372, opened 2023-12-04
fix the issue that the session's Close method could be invoked multi times

bruce1125:
Although the framework uses a sync.map and some check logic to prevent the SYNC issue, there is possibility to trigger in some situation.

### DIFF
--- a/session/session.go
+++ b/session/session.go
@@ -99,6 +99,7 @@ type sessionImpl struct {
 	Subscriptions       []*nats.Subscription                  // subscription created on bind when using nats rpc server
 	requestsInFlight    ReqInFlight                           // whether the session is waiting from a response from a remote
 	pool                *sessionPoolImpl
+	closed              sync.Once
 }
 
 type ReqInFlight struct {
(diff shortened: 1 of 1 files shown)


bruce1125:
There is a possibility that the method is invoked from HandlerService.Handle's defer logic at the same time from different goroutine.
So, we need to make sure two things:
1. ensure the Close method's logic is done only once, for the reason to prevent deleting the new one
2. double check that there isn't a session in the store data before Storing

session/session.go
```diff
@@ -437,6 +438,14 @@ func (s *sessionImpl) Bind(ctx context.Context, uid string) error {
 		if val, ok := s.pool.sessionsByUID.Load(uid); ok {
 			val.(Session).Close()
 		}
+		// although here are checking old session and invoking Close method,
+		// there is a possibility that the method is invoked from HandlerService.Handle's defer logic at the same time from different goroutine
+		// so we need to make sure two things:
+		// 1. ensure the Close method's logic is done only once, for the reason to prevent deleting the new one
+		// 2. double check that there isn't a session in the store data before Storing
+		for s.pool.GetSessionByUID(uid) != nil {
```
felipejfc, line 446: Can we put a timeout here so that this doesn't turn into a dead lock in corner cases?

felipejfc (changes requested):

session/session.go
bruce1125: Sure, That's great. But I am not sure that how much timeout value we could set, is 30s enough? Or set the loop counter limit  directly?

felipejfc: Make it configurable

PR #375, opened 2023-12-13
allow to customize the timeout of rpc

bruce1125:
To customize the timeout value, the only thing we need to do is just like the code below before RPC call and then use the ctx as the context argument:
`
ctx := pcontext.AddToPropagateCtx(context.Background(), constants.RequestTimeout, (60 * time.Second).String())
`

### DIFF
--- a/cluster/nats_rpc_client.go
+++ b/cluster/nats_rpc_client.go
@@ -168,9 +168,13 @@ func (ns *NatsRPCClient) Call(
 		defer session.SetRequestInFlight(requestID, "", false)
 	}
 
-	logger.Log.Debugf("[rpc_client] sending remote nats request for route %s with timeout of %s", route, ns.reqTimeout)
+	reqTimeout := pcontext.GetFromPropagateCtx(ctx, constants.RequestTimeout)
+	if reqTimeout == nil {
+		reqTimeout = ns.reqTimeout.String()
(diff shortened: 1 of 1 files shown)


rsafonseca:
Hey @bruce1125 , what are you trying to achieve exactly? The timeout for the RPCs can already be configured via "pitaya.cluster.rpc.client.grpc.requesttimeout" or "pitaya.cluster.rpc.client.nats.requesttimeout" config options :)

bruce1125:
> Hey @bruce1125 , what are you trying to achieve exactly? The timeout for the RPCs can already be configured via "pitaya.cluster.rpc.client.grpc.requesttimeout" or "pitaya.cluster.rpc.client.nats.requesttimeout" config options :)

Thank you for review. Although we can configure the grpc/nats timeout, that is global. It can cover most of cases excluding some time-consuming operations, such as getting an external API response, collecting a cpu/memory snapshot, etc. And I don't want to set the timeout value only for these issues..

felipejfc (approved):
Sounds like a valid use case. Will be merging it

felipejfc:
@bruce1125 would appreciate if you could update the docs later covering this new capability

bruce1125:
> @bruce1125 would appreciate if you could update the docs later covering this new capability

@felipejfc It's my pleasure,I've created a new [PR](https://github.com/topfreegames/pitaya/pull/393#issue-2260613481).
I am not sure is it OK,and I am realy happy to discuss with you guys if there is any problem.

PR #376, opened 2023-12-15
Added some functions to make it easier to migrate from Pitaya V1

guthyerrz:


session/static.go
```diff
@@ -36,3 +37,16 @@ func OnSessionClose(f func(s Session)) {
 func CloseAll() {
 	DefaultSessionPool.CloseAll()
 }
+
+// New returns a new session instance
+// a NetworkEntity is a low-level network instance
+func New(entity networkentity.NetworkEntity, frontend bool, UID ...string) Session {
+	DefaultSessionPool = NewSessionPool()
```
felipejfc, line 44: Is it intended that you are resetting the DefaultSessionPool every time you create a new session isntance?

logger/interfaces/interfaces.go
```diff
@@ -29,4 +31,6 @@ type Logger interface {
 	WithFields(fields map[string]interface{}) Logger
 	WithField(key string, value interface{}) Logger
 	WithError(err error) Logger
+
+	LogrusLogger() logrus.FieldLogger
```
felipejfc, line 35: Can you explain the need for this method? It sounds a bit weird that we are tying our generic Logger interface to Logrus

felipejfc (changes requested):

session/static.go
guthyerrz: I'll remove this method, It doesn't make sense to be here I just used it for testing on Zooba

logger/interfaces/interfaces.go
guthyerrz: So in a few places I need to get the internal logger that this interface uses, how about this?

`GetInternalLogger() any`

felipejfc: Can't we define the internal logger as another interface rather than using any?

guthyerrz: Like this `GetInternalLogger() logrus.FieldLogger`

felipejfc: no because this interface is too extensive and use internal logrus stuff

logger/interfaces/interfaces.go
```diff
@@ -32,5 +32,5 @@ type Logger interface {
 	WithField(key string, value interface{}) Logger
 	WithError(err error) Logger
 
-	LogrusLogger() logrus.FieldLogger
+	GetInternalLogger() logrus.FieldLogger
```
felipejfc, line 35: change return to any so that we don't restrict ourselves to logrus

felipejfc (changes requested):

felipejfc (approved):

PR #381, opened 2023-12-20
feat: adding base64 request and tls connection

SamuelLucasVM:
### **New Features Added**

- requestB64

Receives a string instead of a json like, request function, and its checks whether if this string is a base64 code. After that, the base64 message is decoded and sent to the server.

- connectTls

Connects to the backend with basic TlsConfig, using the InsecureSkipVerify.

### DIFF
--- a/xk6-pitaya/client.go
+++ b/xk6-pitaya/client.go
@@ -1,6 +1,8 @@
 package pitaya
 
 import (
+	"crypto/tls"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


xk6-pitaya/client.go
```diff
@@ -34,6 +36,26 @@ type Client struct {
...
+// ConnectTls connects to the server with a basic tls config
+// addr is the address of the server to connect to
+func (c *Client) ConnectTls(addr string) error {
+	vuState := c.vu.State()
+
+	if vuState == nil {
+		return errors.New("connecting to a pitaya server in the init context is not supported")
+	}
+
+	config := &tls.Config{GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) { return nil, nil }, InsecureSkipVerify: true}
```
reinaldooli, line 48: I'm missing a way to pass the TLS Certificates.

felipejfc (approved):

PR #390, opened 2024-03-18
feat: adding tls config into connect and base64 request

SamuelLucasVM:


xk6-pitaya/client.go
```diff
@@ -100,6 +107,47 @@ func (c *Client) Notify(route string, msg interface{}) error {
 	return c.client.SendNotify(route, data)
 }
 
+// RequestB64 sends a request to the server using a base64 string
+// route is the route to send the request to
+// str is the string passed in request
+// returns a promise that will be resolved when the response is received
+// the promise will be rejected if the timeout is reached before a response is received
+func (c *Client) RequestB64(route string, b64msg string) *goja.Promise { // TODO: add custom timeout
```
felipejfc, line 115: What is the use case for this method?

SamuelLucasVM: This for some base64 message, in my case used this to pass some encoded protos to my server

PR #393, opened 2024-04-24
Feat/customize rpc timeout

bruce1125:
update docs to explain how to customize the timeout value

### DIFF
--- a/cluster/nats_rpc_client.go
+++ b/cluster/nats_rpc_client.go
@@ -168,9 +168,13 @@ func (ns *NatsRPCClient) Call(
 		defer session.SetRequestInFlight(requestID, "", false)
 	}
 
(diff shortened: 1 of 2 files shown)


felipejfc (approved):
LGTM.
We should create a helper later like, a new method in the interface like:
RPC(msg, route, ..., timeout) which sets this value on the context, so that this feature would be easier accessible and we avoid unmarshal errors on the timeout value

felipejfc:
FYI new version released with this change: https://github.com/topfreegames/pitaya/releases/tag/v2.11.0

bruce1125:
> LGTM. We should create a helper later like, a new method in the interface like: RPC(msg, route, ..., timeout) which sets this value on the context, so that this feature would be easier accessible and we avoid unmarshal errors on the timeout value

That's a good idear, waiting for the new API.

PR #394, opened 2024-04-24
New module type for implementing app level session draining

paulikaila-wl:
Introduces a new interface SessionModule. If a Module implements the SessionModule interface it will be treated differently when session draining is enabled.

When session draining starts, Pitaya will call `StartSessionDraining()` on all registered SessionModules. And when checking if sessions have been drained, Pitaya will call `SessionCount()` on all registered SessionModules to check they have all reached zero.

### DIFF
--- a/app.go
+++ b/app.go
@@ -129,31 +129,32 @@ type Pitaya interface {
 
 // App is the base app struct
 type App struct {
-	acceptors        []acceptor.Acceptor
-	config           config.PitayaConfig
-	debug            bool
-	dieChan          chan bool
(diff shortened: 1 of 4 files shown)


felipejfc:
Thx for the contribution @paulikaila-wl! Do you mind documenting it? here: https://github.com/topfreegames/pitaya/blob/main/docs/features.md#modules

paulikaila-wl:
I added some documentation.

felipejfc (approved):
LGTM

PR #406, opened 2024-06-25
Fix/observability gaps

renatoaf:
- Add tracing span for connection writes (we're noticing some high times in some of our games, we might want to add a timeout here in the future)
- Properly wrap errors when low-level connection writes fail (they are currently failing with a PIT-000 code in the metrics)
- Add error logging on some spots we were previously missing

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -511,7 +511,11 @@ func (a *agentImpl) write() {
 					"conn": remoteAddress,
 				}
 
-				ctx = tracing.StartSpan(ctx, "conn write", tags)
+				parent, err := tracing.ExtractSpan(ctx)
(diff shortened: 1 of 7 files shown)


felipejfc (approved):

felipejfc:
@renatoaf idk if it was with these changes, but the unit tests are failing, I've already re-ran they 3 times but no luck. can you check before merging?

renatoaf:
> @renatoaf idk if it was with these changes, but the unit tests are failing, I've already re-ran they 3 times but no luck. can you check before merging?

Yep. Fixed the test and added a new one covering this scenario (closed agent after a write fails)

PR #416, opened 2024-08-06
chore: only output log in debug when we cannot get next message

kalabiyau:
The following we observe in production:

```
Error reading next available message: read tcp 192.168.1.1:3250->192.168.1.2:34184: read: connection reset by peer
```

These connections are part of the TCP health-checks which AWS LB does. Healthcheck succeed as it can establish a connection successfully, but pitaya would produce this error log. It might be useful to find such cases but in higher log level - e.g. Debug


kalabiyau:
superseded by https://github.com/topfreegames/pitaya/pull/417

PR #423, opened 2024-11-07
feat: add QUIC server support

juniorjse:
Enable support for QUIC protocol to expand connection options in supported environments.


examples/demo/cluster_grpc/main_quic.go
```diff
@@ -0,0 +1,160 @@
+package main
```
felipejfc, line 1: Instead of creating a whole new main, why didn't you just add a flag to the existing main.go?

pkg/acceptor/quic_acceptor.go
```diff
@@ -0,0 +1,244 @@
...
+// Close closes the QUIC listener
+func (a *QuicAcceptor) Close() error {
+	if a.listener != nil { // Correct comparison
+		return a.listener.Close()
+	}
+	return nil
+}
+
+// EnableProxyProtocol not implemented for QUIC, keep as No-op or implement as needed
+func (a *QuicAcceptor) EnableProxyProtocol() {
```
felipejfc, line 111: @guthyerrz I think it's important to understand who's using proxy protocol and what's the use case, so we evaluate if we need to implement this right away or not

pkg/acceptor/quic_acceptor.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return stream.Read(p)
+}
+
+// Write writes data to the connection with a defined deadline
+func (q *QuicConnWrapper) Write(p []byte) (int, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second) // 5 seconds timeout as an example
```
felipejfc, line 172: Make this timeout configurable instead of hardcoding it here

pkg/acceptor/quic_acceptor.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return stream.Write(p)
+}
+
+// Close closes the QUIC connection
+func (q *QuicConnWrapper) Close() error {
+	return q.conn.CloseWithError(0, "closed")
```
felipejfc, line 185: Didn't really read the API here, but do all connections need to be closed with errors?

pkg/acceptor/quic_acceptor.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// RemoteAddr returns the remote address of the connection
+func (q *QuicConnWrapper) RemoteAddr() net.Addr {
+	return q.conn.RemoteAddr()
+}
+
+// GetNextMessage reads the next message available in the QUIC stream
+func (q *QuicConnWrapper) GetNextMessage() (b []byte, err error) {
+	// Read data from the stream
```
felipejfc, line 200: can't you leverage io utilities that go provides in this method, just like we do for the tcp acceptor?
https://github.com/topfreegames/pitaya/blob/main/pkg/acceptor/tcp_acceptor.go#L55-L74

felipejfc: also, you seem to be assuming the maximum size of a message will be 1024 bytes, which is inaccurate

felipejfc (changes requested):

examples/demo/cluster_grpc/main_quic.go
juniorjse: The new main was created to avoid interfering with existing main code and maintain separation of responsibilities between execution modes. But we can easily adjust this.

pkg/acceptor/quic_acceptor.go
juniorjse: I'll do it

pkg/acceptor/quic_acceptor.go
juniorjse: I had initially left the limit fixed as a testing approach, but I tested it with io utilities, and it worked fine. Thanks for pointing that out; it was something I had overlooked

examples/demo/cluster_grpc/main_quic.go
felipejfc: lets adjust it. I rather keep a single one

PR #424, opened 2024-12-11
Fix scenarios where error messages aren't propagated correctly

felippeduran:
Currently, there are three scenarios where pitaya errors aren't propagated correctly:

* When pitaya errors are wrapped: `errors.NewError` doesn't make proper use of `errors.As`, triggering the issue.
* When using a custom serializer: in this case, `utils.GetErrorFromPayload` looks for explicit implementations, breaking polymorphism and inducing the bug.
* When using a json serializer: this was caused by `utils.GetErrorPayload`, that would convert any error to the protobuf version of a pitaya error, which would then fail to get parsed by `utils.GetErrorFromPayload`

For my specific case, I encountered these issues when trying to return pitaya errors in the routing function. The first one would convert it to a generic pitaya error `Internal` before returning to clients, and the other two would make error messages not propagate correctly to metrics, having errors as `Unknown` in Prometheus.

This MR fixes all three scenarios, with the following changes:
* `utils.GetErrorFromPayload` now always assumes error payloads are in protobuf schema. This is aligned with the `utils.GetErrorPayload`, which always serializes with the same schema.
* `errors.NewError` now uses `errors.As` when trying to merge pitaya errors.

I also took the time to improve some test cases, implement a few more and test things against pitaya-cli.

Notes:
An alternative would be to make the serialization methods use a non-protobuf schema for JSON. However, the C# `PitayaClient` always assumes errors to be in the protobuf schema, with `code` and `msg` (see [here](https://github.com/topfreegames/libpitaya/blob/master/unity/PitayaExample/Assets/Pitaya/PitayaBinding.cs#L406-L407)). I don't know if that was intentional, or an accidental behavior due to the previously asymmetric `utils.GetErrorFromPayload` and  `utils.GetErrorPayload`.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -523,6 +523,7 @@ func TestAgentResponseMID(t *testing.T) {
 			expected := pendingWrite{ctx: ctx, data: []byte("ok!"), err: nil}
 			var err error
 			if table.msgErr {
+				mockSerializer.EXPECT().Unmarshal(gomock.Any(), gomock.Any()).Return(nil)
 				err = ag.ResponseMID(ctx, table.mid, table.data, table.msgErr)
 			} else {
 				err = ag.ResponseMID(ctx, table.mid, table.data)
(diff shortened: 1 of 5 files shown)


hspedro (approved):
LGTM

PR #425, opened 2024-12-16
Fix scenarios where error messages aren't propagated correctly (#424)

hspedro:
Porting the routing fix that @felippeduran provided to main/v3. See #424 

* Improve tests for agent and service/remote to expose bad behaviors
* Breakdown test into multiple ones to make them more clear
* Fix scenarios where error messages aren't propagated correctly

---------

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -39,8 +39,6 @@ import (
 	"github.com/topfreegames/pitaya/v3/pkg/logger/interfaces"
 	"github.com/topfreegames/pitaya/v3/pkg/protos"
 	"github.com/topfreegames/pitaya/v3/pkg/serialize"
-	"github.com/topfreegames/pitaya/v3/pkg/serialize/json"
-	"github.com/topfreegames/pitaya/v3/pkg/serialize/protobuf"
 	"github.com/topfreegames/pitaya/v3/pkg/tracing"
 
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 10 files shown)


pkg/util/util.go
felipejfc, line 1: Will GetErrorFromPayload keep working if the server is using json serializer?

felipejfc (commented):
only one question. apart from it, LGTM

pkg/util/util.go
hspedro: Yes. The difference between unmarshalling to `e.Error` struct and `protos.Error` is a few internal XXX fields:

```go
// Error is an error with a code, message and metadata
type Error struct {
	Code     string
	Message  string
	Metadata map[string]string
}
```

```go
type Error struct {
	Code                 string            `protobuf:"bytes,1,opt,name=code" json:"code,omitempty"`
	Msg                  string            `protobuf:"bytes,2,opt,name=msg" json:"msg,omitempty"`
	Metadata             map[string]string `protobuf:"bytes,3,rep,name=metadata" json:"metadata,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
	XXX_NoUnkeyedLiteral struct{}          `json:"-"`
	XXX_unrecognized     []byte            `json:"-"`
	XXX_sizecache        int32             `json:"-"`
}
```

So JSON serializer should still be able to serialize to protos.Error struct, just the XXX fields will be empty

reinaldooli (approved):

PR #432, opened 2025-01-14
feat(nats): add compression config

hspedro:
In NATS client and server, if the connection uses websocket we can enable compression by passing the Compression config - it needs to be enbaled on both client and server conn setup. Thus creating a new config to enbale on both sides - if not websocket, this config has no effect.

Added documentation for both variables as well:

  * - pitaya.cluster.rpc.client.nats.websocketcompression
    - true
    - bool
    - Enables compression in websocket connections to NATS. Needs both client and server to be enabled

  * - pitaya.cluster.rpc.server.nats.websocketcompression
    - true
    - bool
    - Enables compression in websocket connections to NATS. Needs both client and server to be enabled

NATS Compression lib ref: https://pkg.go.dev/github.com/nats-io/nats.go#Compression


rsafonseca (approved):
LGTM :)

PR #433, opened 2025-01-14
Port connection info changes to v2

hspedro:
Couple of changes were added to v2.9.1, v2.9.2 and v2.9.3 that we need to port to the mainline of v2:
* #430 
* #431 

## Summary

Get number of connected clients and expose this via API


reinaldooli (approved):

rsafonseca (approved):
LGTM :)

PR #436, opened 2025-01-17
Add support to loopback RPCs

felippeduran:
The current implementation doesn't support RPC calls to the same server or server type. Although it's nice to block this functionality to avoid developer error, there are some scenarios where it could be interesting to have those. One example is to isolate features in a monolithic server to then migrate later to separate services more easily.

One downside of there RPCs is the unnecessary overhead. In order to avoid such problem, this merge request creates a loopback flow inside the RemoteService class, bypassing nats or gRPC implementations and calling the remote handlers directly. In order to keep an intuitive trace, it also mimics the span structure of regular RPCs.

Tests have been fixed/edited to incorporate the new use case.


cluster/server.go
```diff
@@ -34,6 +34,7 @@ type Server struct {
 	Metadata map[string]string `json:"metadata"`
 	Frontend bool              `json:"frontend"`
 	Hostname string            `json:"hostname"`
+	Loopback bool              `json:"loopback"`
```
felippeduran, line 37: Have you considered naming the field `AllowLoopback` instead? I'm thinking that having it called only `Loopback` could give the impression that it'd loopback all calls, but maybe I'm overthinking it!

felippeduran (commented):
I made some small suggestions, but things look good 👍

I can't approve the MR being the author, though. 😄

config/viper_config.go
```diff
@@ -73,6 +73,7 @@ func (c *Config) fillDefaultValues() {
 		"pitaya.cluster.rpc.client.nats.pinginterval":           pitayaConfig.Cluster.RPC.Client.Nats.PingInterval,
 		"pitaya.cluster.rpc.client.nats.maxpingsoutstanding":    pitayaConfig.Cluster.RPC.Client.Nats.MaxPingsOutstanding,
 		"pitaya.cluster.rpc.client.nats.requesttimeout":         pitayaConfig.Cluster.RPC.Client.Nats.RequestTimeout,
+		"pitaya.cluster.rpc.server.loopbackenabled":             pitayaConfig.Cluster.RPC.Server.LoopbackEnabled,
```
hspedro, line 76: can you add documentation in configuration.rst for that new config and its default value?

hspedro (approved):
LGTM 🚀

PR #437, opened 2025-01-21
Nats connection options

rsafonseca:


cluster/nats_rpc_common.go
```diff
@@ -34,8 +34,8 @@ func getChannel(serverType, serverID string) string {
 func setupNatsConn(connectString string, appDieChan chan bool, options ...nats.Option) (*nats.Conn, error) {
 	natsOptions := append(
 		options,
-		nats.DisconnectHandler(func(_ *nats.Conn) {
-			logger.Log.Warn("disconnected from nats!")
+		nats.DisconnectErrHandler(func(_ *nats.Conn, err error) {
+			logger.Log.Warnf("disconnected from nats! Reason: %q\n", err)
```
hspedro, line 38: Why we're using `%q`, this would translate to a single-quoted character, don't we want to go with `%s`?

config/viper_config.go
```diff
@@ -67,12 +67,22 @@ func (c *Config) fillDefaultValues() {
 		"pitaya.cluster.rpc.client.nats.connectiontimeout":      pitayaConfig.Cluster.RPC.Client.Nats.ConnectionTimeout,
 		"pitaya.cluster.rpc.client.nats.maxreconnectionretries": pitayaConfig.Cluster.RPC.Client.Nats.MaxReconnectionRetries,
 		"pitaya.cluster.rpc.client.nats.websocketcompression":   pitayaConfig.Cluster.RPC.Client.Nats.WebsocketCompression,
+		"pitaya.cluster.rpc.client.nats.reconnectjitter":        pitayaConfig.Cluster.RPC.Client.Nats.ReconnectJitter,
+		"pitaya.cluster.rpc.client.nats.reconnectjittertls":     pitayaConfig.Cluster.RPC.Client.Nats.ReconnectJitterTLS,
+		"pitaya.cluster.rpc.client.nats.reconnectwait":          pitayaConfig.Cluster.RPC.Client.Nats.ReconnectWait,
+		"pitaya.cluster.rpc.client.nats.pinginterval":           pitayaConfig.Cluster.RPC.Client.Nats.PingInterval,
+		"pitaya.cluster.rpc.client.nats.maxpingsoutstanding":    pitayaConfig.Cluster.RPC.Client.Nats.MaxPingsOutstanding,
```
hspedro, line 74: Can we add to `configuration.rst` document as well?

hspedro (approved):
LGTM with minor suggestions

cluster/nats_rpc_common.go
rsafonseca: %q will quote each element of a slice of strings not a single character
i used %q because that's what was used in the nats.go examples

config/viper_config.go
rsafonseca: good catch, done!

PR #439, opened 2025-01-23
fix(nats): wait for reconnects on setup

hspedro:
If the initial connect fails, nats will spawn reconnect async handlers. Thus, we need to wait for all reconnects to be attempted before returning to the caller, otherwise, we won't be making use of reconnections

Ref: https://github.com/nats-io/nats.go/blob/bb3ad1c6282d4e710dd3d78171763c10ce41be53/test/conn_test.go#L1096-L1112

### DIFF
--- a/cluster/nats_rpc_common.go
+++ b/cluster/nats_rpc_common.go
@@ -22,6 +22,9 @@ package cluster
 
 import (
 	"fmt"
+	"os"
+	"syscall"
+	"time"
 
 	nats "github.com/nats-io/nats.go"
 	"github.com/topfreegames/pitaya/v2/logger"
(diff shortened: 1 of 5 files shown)


cluster/nats_rpc_common.go
```diff
@@ -61,11 +62,20 @@ func setupNatsConn(connectString string, appDieChan chan bool, options ...nats.O
 				logger.Log.Errorf(err.Error())
 			}
 		}),
+		nats.ConnectHandler(func(*nats.Conn) {
```
reinaldooli, line 65: Will this `ConnectHandler` only execute the given function if the connection succeeds?

hspedro: Yes

cluster/nats_rpc_common_test.go
```diff
@@ -77,3 +78,36 @@ func TestNatsRPCCommonCloseHandler(t *testing.T) {
 	assert.True(t, ok)
 	assert.True(t, value)
 }
+
+func TestSetupNatsConnReconnection(t *testing.T) {
```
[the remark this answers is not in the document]
hspedro: Added

reinaldooli (approved):

PR #443, opened 2025-02-07
fix(nats): drain connections upon error

hspedro:
When failing to connect, drain connection rather than closing directly. This is safer than calling close and will wait for connections before closing, gracefully handling ongoing reconnects and avoid leaking goroutines in test


rsafonseca (approved):
LGTM

PR #447, opened 2025-03-24
Fix dieChan not getting closed or pitaya not terminating on fatal etcd and nats failures

felippeduran:
This issue could cause pitaya to get into a zombie state in case of fatal etcd or nats failure.

Some applications usually listen to this channel with `GetDieChan()` in order to handle pitaya's termination. In these cases, pitaya would run its termination flow by disposing modules and components, but the channel would never return for the application, leading to the zombie state.

Another problem is that using `GetDieChan()` and listening to it in the application side can swallow the internal termination message. This in turn leads to pitaya never entering the termination flow and the application getting stuck waiting for `Start()` to return.

Example of fragile usage that can trigger both issues currently ❌:
```go
func (a *App) Start() {
	dieChan := a.pitaya.GetDieChan()

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		a.pitaya.Start()
	}()

        // wait for pitaya to start.

	fmt.Println("server started")

	<-dieChan // <---- termination might get stuck here waiting for the channel to get closed

	fmt.Println("server is terminating")

	wg.Wait() // <---- or if we get the die message from dieChan above, then we get stuck here waiting for a.pitaya.Start() to finish
}
```

The proposed pull request addresses both cases.

However, in order to fix the second one, a small behavioral API change was required (see second commit). Now, we publicly expose a proxy, one-way, channel in `GetDieChan()` that the application layer can listen to. We can't really make it two-way as we might face the same problem again.

The alternative in order to keep the public API completely untouched would be to either deprecate `GetDieChan()` completely or alway require a call to `Shutdown()`, even though pitaya might be terminating already. Here's how each approach could look like:

Using `Shutdown()`:
```go
func (a *App) Start() {
	dieChan := a.pitaya.GetDieChan()

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		a.pitaya.Start()
	}()

        // wait for pitaya to start.

	fmt.Println("server started")

	<-dieChan

	a.pitaya.Shutdown() // <---- invoking Shutdown() here covers the edge case of the previous line swallowing the termination signal and preventing Start() to return

	fmt.Println("server is terminating")

	wg.Wait()
}
```

Without relying on `GetDieChan()` at all:
```go
func (a *App) Start() {
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		defer wg.Done()
		a.pitaya.Start()
	}()

        // wait for pitaya to start.

	fmt.Println("server started")

	wg.Wait()

	fmt.Println("pitaya terminated")
}
```

And as a final note, it might be worth mentioning that my recommendation for `v3` is to remove `GetDieChan()` completely and solely rely on the last example as good API usage.

### DIFF
--- a/app.go
+++ b/app.go
@@ -70,8 +70,7 @@ const (
 
 // Pitaya App interface
 type Pitaya interface {
-	// GetDieChan gets the channel that the app sinalizes when its going to die. Note that listening to this channel
-	// might swallow internal signals, so it's recommended to wait for Start to return or invoke Shutdown when terminating the application.
+	// GetDieChan gets the channel that the app sinalizes when its going to die.
 	GetDieChan() chan bool
 	SetDebug(debug bool)
 	SetHeartbeatTime(interval time.Duration)
@@ -159,6 +158,8 @@ type App struct {
 	sessionModulesArr []sessionModuleWrapper
 	groups            groups.GroupService
 	sessionPool       session.SessionPool
+	externalDieChan   chan bool
+	sgChan            chan os.Signal
 }
 
 // NewApp is the base constructor for a pitaya app instance
@@ -205,6 +206,8 @@ func NewApp(
 		modulesArr:        []moduleWrapper{},
 		sessionModulesArr: []sessionModuleWrapper{},
 		sessionPool:       sessionPool,
+		externalDieChan:   make(chan bool),
+		sgChan:            make(chan os.Signal, 1),
 	}
 	if app.heartbeat == time.Duration(0) {
 		app.heartbeat = config.Heartbeat.Interval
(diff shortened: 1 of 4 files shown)


app.go
```diff
@@ -367,9 +371,12 @@ func (app *App) Start() {
 				}
 			}
 		}
-		close(app.dieChan)
```
joaobologna, line 370: Will `dieChan` be closed by `app.Shutdown()` now?

felippeduran: Yes, calling `app.Shutdown()` is safer because it handles an already closed channel. Currently, if the code calls `app.Shutdown()` from outside pitaya at the same time it begins to terminate due to a kill signal, for example, we might get a panic when trying to close the channel twice depending on whoever does it first.

joaobologna (approved):
LGTM

PR #454, opened 2025-05-16
Fix sessionByUID leak when there's a race condition between session closing and uid binding

felippeduran:
It might happen that the server processes a `uid` binding at the same time, or right after the session is closed. In such situations, the session is re-added to the `sessionByUID`, making it leak and be wrongly accessible when calling `GetSessionByUID`.

Once this happens, it ends up making the issue reported by #452 happen consistently for users when trying to connect to the same instance that had the session leaked. For the scenario of multiple frontend instances, the problem is mitigated, as the chances of a client ending up consistently in the same instance is small.

This PR addresses the issue by validating whether the session is present in `sessionByID` before binding the `uid`. The validation is done before the `sessionBindCallbacks` are called so to avoid triggering false positive invocations.

### DIFF
--- a/e2e/e2e_test.go
+++ b/e2e/e2e_test.go
@@ -223,6 +223,7 @@ func TestUniqueSession_ShouldSucceedIfKickFails(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Force break the connection so that the kick fails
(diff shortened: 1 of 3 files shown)


felippeduran:
Note that this PR depends on #452 to be merged first.

PR #494, opened 2026-04-24
fix(agent): capture error code before packetEncodeMessage mutates payload

victordomiciano:
## Problem

`MessagesEncoder.Encode` mutates `message.Data` in place when compression
produces smaller output than the source. The `send()` function in
`agent/agent.go` reads the error code from `m.Data` **after** calling
`packetEncodeMessage`, so when compression triggers, `util.GetErrorFromPayload`
receives deflated bytes. The JSON unmarshal inside that function silently
fails (the error is discarded with `_ = serializer.Unmarshal(...)`), and
the returned `Error` falls back to `ErrUnknownCode` (`"PIT-000"`).

## Effect

The wire payload still carries the correct code — **clients are not affected**.
But `pendingWrite.err`, and therefore the `code` tag on the `response_time`
metric emitted by `ReportTimingFromCtx`, loses the code for any error whose
JSON payload compresses smaller than the source (typically any error that
carries metadata via `pitaya.Error(err, code, metadata)`).

Monitors that filter by `code` (e.g. excluding expected application errors
from an error rate alert) break silently because the tag they expect is
replaced with `PIT-000`. Error kinds become indistinguishable in metrics.

The bug is payload-size dependent: small errors (no metadata, deflate
overhead > savings) are unaffected.

## Fix

Capture the error via `util.GetErrorFromPayload` **before**
`packetEncodeMessage` runs, while `m.Data` still holds the original
serialized bytes. Only the metric tag is affected; the wire format, packet
layout, and payload sent to clients are unchanged.

## Test

Adds `TestAgentResponseMIDPreservesErrorCodeWithCompression`. Builds an
error payload large enough to trigger compression, calls `ResponseMID`
with `isError=true` through an agent configured with a real JSON serializer
and a `MessagesEncoder(true)`, asserts `pendingWrite.err` carries the
original code.

Without this change the test fails with:
pendingWrite.err.Code was "PIT-000", expected "PIT-COMPRESSION-TEST"

## Scope

- Single-function reorder in `send()`.
- No API change.
- No wire format change.
- No behavioral change for non-error responses or for errors whose payload
  does not compress smaller than the source.

---

> [!NOTE]
> **Low Risk**
> Low risk: a small reorder in `agent.send` affecting only error metric tagging, plus a regression test; no protocol or API changes.
> 
> **Overview**
> Fixes a compression edge case where `agent.send` could lose the original error code used for the `response_time` metric tag because `MessagesEncoder.Encode` mutates `message.Data` in-place; the error is now extracted from the payload *before* encoding/compression runs.
> 
> Adds `TestAgentResponseMIDPreservesErrorCodeWithCompression` to reproduce the bug end-to-end with compression enabled and assert `pendingWrite.err` (and thus metric tagging) retains the handler’s original error code.
> 
> Cleans up CI/coverage targets in `Makefile` by pinning `goveralls` to `v0.0.12` and removing profile-merging/`gocovmerge` usage in favor of `coverprofile.out` directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5880a3a38c5a497cfa9ee466740fa0b8bf7babab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,7 @@ init-submodules:
 	@git submodule init
 
 setup-ci:
-	@go install github.com/mattn/goveralls@latest
-	@go install github.com/wadey/gocovmerge@latest
+	@go install github.com/mattn/goveralls@v0.0.12
 
 setup-protobuf-macos:
 	@brew install protobuf
@@ -165,16 +164,12 @@ test-coverage: unit-test-coverage
 test-coverage-html: test-coverage
 	@go tool cover -html=coverprofile.out
 
-merge-profiles:
-	@rm -f coverage-all.out
-	@gocovmerge *.out > coverage-all.out
-
-test-coverage-func coverage-func: test-coverage merge-profiles
+test-coverage-func coverage-func: test-coverage
 	@echo
 	@echo "\033[1;34m=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-\033[0m"
 	@echo "\033[1;34mFunctions NOT COVERED by Tests\033[0m"
 	@echo "\033[1;34m=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-\033[0m"
-	@go tool cover -func=coverage-all.out | egrep -v "100.0[%]"
+	@go tool cover -func=coverprofile.out | egrep -v "100.0[%]"
 
 mocks: agent-mock session-mock networkentity-mock pitaya-mock serializer-mock metrics-mock acceptor-mock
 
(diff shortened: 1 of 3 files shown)


victordomiciano:
The Unit Test failure is caused by `golang.org/x/tools v0.44.0` (released 2026-04-09) requiring Go 1.25, while the v2 CI runs on Go 1.24. The `setup-ci` Makefile target uses `@latest` for tooling, so this affects any v2 CI run after April 9 regardless of the diff. Repro: rerun the workflow on the previous commit (7343bca) and it will fail the same way. Happy to bump Go to 1.25 in this PR if maintainers prefer.

gmurayama (approved):

